PR #33, opened 2024-04-14
[RELEASE-V8] Add new fields

StevenRenaux:
Add new form fields failOnHttpStatusCodes 
Add new form fields skipNetworkIdleEvent 
Add new form fields cookies


Neirda24 (commented):
What about instead of passing by an associative array we type with the cookie object from symfony directly?

Neirda24 (approved):

Neirda24 (approved):

PR #77, opened 2024-06-21
[UPDATE] Fixes configuration + a few stuff

Neirda24:


StevenRenaux (approved):

src/DependencyInjection/CompilerPass/GotenbergPass.php
```diff
@@ -16,7 +16,7 @@ public function process(ContainerBuilder $container): void
         }
 
         foreach ($container->findTaggedServiceIds('sensiolabs_gotenberg.pdf_builder') as $serviceId => $tags) {
-            $container->register('debug.'.$serviceId, TraceablePdfBuilder::class)
+            $container->register('.debug'.$serviceId, TraceablePdfBuilder::class)
```
Jean-Beru, line 19: It seems that a dot is missing after `debug`.
```suggestion
            $container->register('.debug.'.$serviceId, TraceablePdfBuilder::class)
```

Neirda24: because the services we define are prefixed with a `.` I made it disappear here to avoid the `..` when doing debug:container. Still a thing to check though when using a custom service because then it is missing one indeed

Jean-Beru: OK, I didn't notice that these services were already prefixed.

PR #84, opened 2024-06-25
[Feature] Async generation

### LINKED ISSUE #46: Implements the Webhook features
Using Webhook & RemoteEvent components ?

maelanleborgne:
Closes #46 

Gotenberg allows async generation of files by passing webhooks urls in dedicated headers. Doing this, the request to Gotenberg server returns a 204 immediately, and once the file is generated, a call will be made by Gotenberg server on the specified endpoint to upload the generated file. This PR aims at enabling easy integration of this feature with the bundle.

The base of the feature is a new interface `AsyncBuilderInterface` that defines a `generateAsync` method. The `AbstractPdfBuilder` and `AbstractScreenshotBuilder` now implement this interface.
Another important addition is the `webhookUrls(string $successWebhook, string|null $errorWebhook)` that allows directly setting the webhooks url to use with the builders.

To make things easier when not working directly with urls (if you'd like to generate the url from a route, or if using the Webhook component), a some new configuration options are available :
```yaml
# packages/sensiolabs_gotenberg.yaml

sensiolabs_gotenberg:  
    # ...
    webhook:  
        # Here you can define any number of webhook configuration so that you can reuse them in builder configs or to switch configuration at runtime
        my_config_name:  
            success:  
                url: 'http://example.com/webhook'
                # route: ['my_route_name', ['param1', 'param2']]
            # This "error" entry is optional. If not set, the "success" configuration will be used for both success and error.
            error:  
                url: 'http://example.com/webhook-error'
                # route: ['my_route_name', ['param1', 'param2']]
    default_options:
        # Here you can define which webhook config to use by default
        webhook: 'my_config_name'
        pdf:
            html:
                webhook: 'my_config_name'
                # Or you can define a specific configuration directly from here
                # webhook:
                #    success:
                #	    url: 'http://example.com/custom-webhook'
    #...
```

TODO:
- [x] Write documentation
- [ ] Add example config to flex recipe
- [x] Change icon in profiler
- [ ] Add default webhook configuration in profiler
- [x] Add/fix tests


Neirda24:
After discussing this with @ConstantBqt we have another usecase to take into account : If we generate the PDF but the webhook is not setup in our own app but is external. We should be able to declare such case. WDYT @maelanleborgne ?

maelanleborgne:
> After discussing this with @ConstantBqt we have another usecase to take into account : If we generate the PDF but the webhook is not setup in our own app but is external. We should be able to declare such case. WDYT @maelanleborgne ?

This is already present in the PR : 

```php
// ...
$this->gotenberg->html()
  ->webhookUrls('https://external.service/webhook', 'https://external.service/webhook-error')
  ->webhookExtraHeaders(['X-ExternalService-Secret' => 'a_secret'])
  ->generateAsync()
;
```

or via the configuration :
```yaml
sensiolabs_gotenberg:  
    # ...
    webhook:  
        external_service:  
            success:  
                url: 'https://external.service/webhook'
            error:  
                url: 'https://external.service/webhook-error'
    default_options:
        pdf:
            html:
                webhook: 'external_service'
```

Or maybe I'm missing the point.

Jean-Beru:
> This is already present in the PR :

Could this part be done in a dedicated PR ?
* allow defining a webhook to add as a header
* make "internal" webhook easier to developer using the component

src/Builder/Pdf/AbstractPdfBuilder.php
```diff
@@ -124,6 +136,63 @@ public function generate(): GotenbergResponse
...
+        }
+        $this->gotenbergClient->call($this->getEndpoint(), $this->getMultipartFormData(), $headers);
+
+        return $operationId;
+    }
+
+    public function webhookConfiguration(string $webhook): static
+    {
+        if (null === $this->webhookConfigurationRegistry) {
+            throw new \LogicException('The WebhookConfigurationRegistry is not available.');
```
Jean-Beru, line 164: We should throw an exception from the Sensiolabs\GotenbergBundle namespace

src/DependencyInjection/Configuration.php
```diff
@@ -36,9 +36,13 @@ public function getConfigTreeBuilder(): TreeBuilder
                         ->end()
                     ->end()
                 ->end()
+                ->append($this->addNamedWebhookDefinition())
                 ->arrayNode('default_options')
                     ->addDefaultsIfNotSet()
                     ->children()
+                        ->scalarNode('webhook')
+                            ->info('Webhook configuration name.')
```
Jean-Beru, line 44: Not this PR I think

src/Builder/Pdf/AbstractPdfBuilder.php
```diff
@@ -29,10 +31,18 @@ abstract class AbstractPdfBuilder implements PdfBuilderInterface
...
+    /**
+     * @var array<string, mixed>
+     */
+    private array $webhookExtraHeaders = [];
+    private \Closure $operationIdGenerator;
 
     public function __construct(
         protected readonly GotenbergClientInterface $gotenbergClient,
         protected readonly AssetBaseDirFormatter $asset,
+        protected readonly WebhookConfigurationRegistry|null $webhookConfigurationRegistry = null,
```
Jean-Beru, line 45: What is the purpose of the `WebhookConfigurationRegistry` ? Could it be handled like other parameters ?

config/builder_pdf.php
```diff
@@ -22,6 +23,7 @@
             service('sensiolabs_gotenberg.client'),
             service('sensiolabs_gotenberg.asset.base_dir_formatter'),
             service('twig')->nullOnInvalid(),
+            service(WebhookConfigurationRegistryInterface::class)->nullOnInvalid(),
```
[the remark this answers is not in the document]
Neirda24: I agree with @Jean-Beru . My remark was not regarding the line itself but let's discuss this together to avoid any misunderstanding.

src/DependencyInjection/Configuration.php
maelanleborgne: This is not related to the webhook component, it's a reference to one of the webhooks configs defined by the user in `sensiolabs_gotenberg.webhook`

src/Builder/Pdf/AbstractPdfBuilder.php
maelanleborgne: It holds the different webhook configs available, so changes can be made at runtime using `->webhookConfiguration('custom_config')`. I'm not sure we can make it work with `setConfiguration` because the config of a builder may be one of : 
- the name of a webhook config defined above in the config file (so it will still have to access the config registry)
- a new definition : success and error with a route/url each (and then we'd have to make the router available in order to generate urls for the routes)

config/builder_pdf.php
```diff
@@ -38,6 +39,7 @@
             service('router')->nullOnInvalid(),
         ])
         ->call('setLogger', [service('logger')->nullOnInvalid()])
+        ->call('setWebhookConfigurationRegistry', [service('.sensiolabs_gotenberg.webhook_configuration_registry')->nullOnInvalid()])
```
Neirda24, line 42: Why through a setter ? Why not the constructor ?

src/Builder/AsyncBuilderTrait.php
```diff
@@ -0,0 +1,88 @@
...
+        ]);
+
+        $this->webhookExtraHeaders['X-Gotenberg-Operation-Id'] = $operationId;
+        $headers = [
+            'Gotenberg-Webhook-Url' => $this->webhookUrl,
+            'Gotenberg-Webhook-Error-Url' => $this->errorWebhookUrl,
+            'Gotenberg-Webhook-Extra-Http-Headers' => json_encode($this->webhookExtraHeaders, \JSON_THROW_ON_ERROR),
+        ];
+        if (null !== $this->fileName) {
+            $headers['Gotenberg-Output-Filename'] = basename($this->fileName, '.pdf');
```
Neirda24, line 35: remove the basename call for now. it will be handled correctly in another PR. Plus : the async is not only for PDFs.

src/Builder/AsyncBuilderTrait.php
```diff
@@ -0,0 +1,88 @@
...
+     * @param array<string, mixed> $extraHeaders
+     */
+    public function webhookExtraHeaders(array $extraHeaders): static
+    {
+        $this->webhookExtraHeaders = array_merge($this->webhookExtraHeaders, $extraHeaders);
+
+        return $this;
+    }
+
+    public function operationIdGenerator(\Closure $operationIdGenerator): static
```
Neirda24, line 77: phpstan annotation for closure please.

src/Builder/AsyncBuilderTrait.php
```diff
@@ -0,0 +1,88 @@
...
+
+        return $this;
+    }
+
+    /**
+     * @param array<string, mixed> $extraHeaders
+     */
+    public function webhookExtraHeaders(array $extraHeaders): static
+    {
+        $this->webhookExtraHeaders = array_merge($this->webhookExtraHeaders, $extraHeaders);
```
Neirda24, line 72: so how do we reset this ?

src/Builder/AsyncBuilderTrait.php
```diff
@@ -0,0 +1,88 @@
...
+    public function operationIdGenerator(\Closure $operationIdGenerator): static
+    {
+        $this->operationIdGenerator = $operationIdGenerator;
+
+        return $this;
+    }
+
+    protected static function defaultOperationIdGenerator(): string
+    {
+        return 'gotenberg_'.bin2hex(random_bytes(16)).microtime(true);
```
Neirda24, line 86: Maybe we should keep the id stateful per builder ? Reset it only when generating the PDF. Also the ID should be part of any operation I guess. Not only the async one. WDYT ?

src/Builder/Pdf/AbstractChromiumPdfBuilder.php
```diff
@@ -586,6 +586,7 @@ protected function addConfiguration(string $configurationName, mixed $value): vo
             'fail_on_console_exceptions' => $this->failOnConsoleExceptions($value),
             'skip_network_idle_event' => $this->skipNetworkIdleEvent($value),
             'metadata' => $this->metadata($value),
+            'webhook' => null,
```
Neirda24, line 589: why ?

src/Builder/AsyncBuilderTrait.php
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    public function webhookConfiguration(string $webhook): static
+    {
+        if (null === $this->webhookConfigurationRegistry) {
+            throw new WebhookConfigurationException('The WebhookConfigurationRegistry is not available.');
+        }
+        $webhookConfiguration = $this->webhookConfigurationRegistry->get($webhook);
+
+        return $this->webhookUrls($webhookConfiguration['success'], $webhookConfiguration['error']);
```
Neirda24, line 56: maybe it can also contain extra headers ? Would that be relevant ? Or authent or anything else ?

src/Debug/Builder/TraceableScreenshotBuilder.php
```diff
@@ -24,7 +25,7 @@ final class TraceableScreenshotBuilder implements ScreenshotBuilderInterface
 
     public function __construct(
         private readonly ScreenshotBuilderInterface $inner,
-        private readonly Stopwatch $stopwatch,
+        private readonly Stopwatch|null $stopwatch,
```
Neirda24, line 28: why this change ?

src/DependencyInjection/Configuration.php
```diff
@@ -651,4 +661,93 @@ private function addPdfMetadata(): NodeDefinition
...
+        $treeBuilder = new TreeBuilder('webhook');
+
+        return $treeBuilder->getRootNode()
+                ->defaultValue([])
+                ->useAttributeAsKey('name')
+                ->arrayPrototype()
+                    ->children()
+                        ->scalarNode('name')
+                            ->validate()
+                                ->ifTrue(static function ($option) {
```
Neirda24, line 676: return type + arg type

src/DependencyInjection/Configuration.php
```diff
@@ -651,4 +661,93 @@ private function addPdfMetadata(): NodeDefinition
...
+                                    return !\is_string($option);
+                                })
+                                ->thenInvalid('Invalid header name %s')
+                            ->end()
+                        ->end()
+                        ->append($this->addWebhookConfigurationNode('success'))
+                        ->append($this->addWebhookConfigurationNode('error'))
+                    ->end()
+                    ->validate()
+                        ->ifTrue(static function ($option): bool {
```
Neirda24, line 686: arg type

src/Builder/AsyncBuilderTrait.php
```diff
@@ -0,0 +1,81 @@
...
+            $result = $result->webhookExtraHeaders($webhookConfiguration['extra_http_headers']);
+        }
+
+        return $result;
+    }
+
+    /**
+     * Allows to set both $successWebhook and $errorWebhook URLs. If $errorWebhook is not provided, it will fallback to $successWebhook one.
+     */
+    public function webhookUrls(string $successWebhook, string|null $errorWebhook = null): static
```
Jean-Beru, line 62: What about adding a $method argument or setter ?

Jean-Beru (commented):
Quick review

src/Debug/Builder/TraceablePdfBuilder.php
```diff
@@ -50,6 +51,30 @@ public function generate(): GotenbergFileResult
...
+        $swEvent = $this->stopwatch?->start($name, 'gotenberg.generate_pdf');
+        $this->inner->generateAsync();
+        $swEvent?->stop();
+
+        $this->pdfs[] = [
+            'calls' => $this->calls,
+            'time' => $swEvent?->getDuration(),
+            'memory' => $swEvent?->getMemory(),
+            'size' => null,
+            'fileName' => null,
```
Jean-Beru, line 72: Maybe we can add a mention/icon to distinguish sync and async calls in the profiler?

Jean-Beru (approved):

Jean-Beru (approved):

Issue #76, opened 2024-06-20
Allow empty array to disable options like failOnHttpStatusCodes


Jean-Beru:
So, I can't disable it. Maybe it will be interesting to allow `[]` as a valid value and `null` to use default.

_Originally posted by @Jean-Beru in https://github.com/sensiolabs/GotenbergBundle/pull/49#discussion_r1647548763_

Neirda24:
Maybe set to `null` or a method `unset(string $key)` WDYT ?

Jean-Beru:
IMO, `failOnHttpStatusCodes` should consider an empty array as a correct value:
- `null` => default http status codes
- `[]` => no http status codes
- `[...]` =>selected status codes

Jean-Beru:
Sure! If Gotenberg doesn't care, this issue is useless :-/

Neirda24:
Just tried it. It works. So yes : 
* `null` should unset the value from being sent.
* `[]` already works and overrides the default so it allows pages like `404` to render as PDF.

Checking on configuration if it is possible.

PR #88, opened 2024-07-01
[UPDATE] Make empty array allowed in some cases

Neirda24:
closes https://github.com/sensiolabs/GotenbergBundle/issues/76


src/Builder/Pdf/AbstractChromiumPdfBuilder.php
```diff
@@ -331,6 +331,10 @@ public function emulatedMediaType(EmulatedMediaType $mediaType): static
      */
     public function cookies(array $cookies): static
     {
+        if ([] === $cookies) {
```
Jean-Beru, line 334: So, it won't be possible to reset cookies ?

Neirda24: my bad. Will add the missing code

Neirda24: done

Jean-Beru (approved):

Issue #50, opened 2024-05-28
Rework the profiler for screenshots


Neirda24:

Jean-Beru:
Closed by #92

PR #92, opened 2024-07-08
Feat: Rework the profiler for screenshots

ConstantBqt:
Rework the profiler to have a single tab with all the generated files (Pdfs and screenshots).

All file lines are clickable

Screenshot from 2024-07-08 09-37-20

All file lines are clickable to display more details
Screenshot from 2024-07-08 09-37-27

We can show or hide the method calls and  the context 
Screenshot from 2024-07-08 09-37-37

Issue https://github.com/sensiolabs/GotenbergBundle/issues/50


Jean-Beru:
Great job! Maybe we could display more information by default. I'm thinking of something like this:
image

src/DataCollector/GotenbergDataCollector.php
```diff
@@ -109,24 +135,23 @@ private function formatSize(int $size): array
...
-     *         'memory': int,
-     *         'size': int<0, max>|null,
-     *         'fileName': string,
-     *         'calls': list<array{
-     *             'method': string,
-     *             'stub': Data
-     *         }>
-     *     }>
+     * @return list<array{
+     *      'builderClass': string,
```
Jean-Beru, line 139: `'` can be removed around keys.

src/DataCollector/GotenbergDataCollector.php
```diff
@@ -60,34 +67,53 @@ public function getName(): string
...
+        $this->lateCollectFiles($this->traceableGotenbergScreenshot->getBuilders(), 'screenshot');
+    }
+
+    /**
+     * @param list<array{string, TraceablePdfBuilder|TraceableScreenshotBuilder}> $builders
+     */
+    private function lateCollectFiles(array $builders, string $type): void
+    {
+        foreach ($builders as [$id, $builder]) {
+            if ($builder->getInner() instanceof AbstractPdfBuilder) {
```
Jean-Beru, line 80: AFAIK, screenshot builder also have the `getMultipartFormData` method

src/DataCollector/GotenbergDataCollector.php
```diff
@@ -60,34 +67,53 @@ public function getName(): string
...
+                dump($builder->getInner()->getMultipartFormData());
```
Jean-Beru, line 81: Not sure that this `dump` is needed

ConstantBqt:
Updated display : 
Screenshot from 2024-07-08 11-24-55

Screenshot from 2024-07-08 11-25-05

src/DataCollector/GotenbergDataCollector.php
```diff
@@ -60,34 +66,49 @@ public function getName(): string
...
+     */
+    private function lateCollectFiles(array $builders, string $type): void
+    {
+        foreach ($builders as [$id, $builder]) {
+            foreach ($builder->getFiles() as $request) {
+                $this->data['request_total_time'] += $request['time'];
+                $this->data['request_total_memory'] += $request['memory'];
+                $this->data['request_total_size'] += $request['size'] ?? 0;
+                $this->data['files'][] = [
+                    'builderClass' => $builder->getInner()::class,
```
Neirda24, line 84: would be great to also the service ID (if it differs from the FQCN that is)

src/DataCollector/GotenbergDataCollector.php
```diff
@@ -60,34 +66,49 @@ public function getName(): string
...
+
+                        return [
+                            'method' => $call['method'],
+                            'stub' => $this->cloneVar(new ArgsStub($args, $call['method'], $call['class'])),
+                        ];
+                    }, $request['calls']),
+                ];
+            }
+
+            $this->data['request_count'] += \count($builder->getFiles());
```
Neirda24, line 111: could be in the foreach ? (avoir double call to getFiles() )

templates/Collector/sensiolabs_gotenberg.html.twig
```diff
@@ -62,13 +25,14 @@
...
+            <span class="sf-toolbar-value" style="margin-left: 4px"> {{ collector.files|length }} </span>
+            <span class="sf-toolbar-label"> files </span>
         {% endset %}
 
         {% set text %}
             <div class="sf-toolbar-info-piece">
-                <strong>PDFs Generated</strong>
-                <span>{{ collector.requestCount }}</span>
+                <strong>Files Generated</strong>
+                <span>{{ collector.files|length }}</span>
```
Neirda24, line 35: the if references requestCount. Seems redundant. could we keep only one of those ?

templates/Collector/sensiolabs_gotenberg.html.twig
```diff
@@ -77,126 +41,145 @@
...
+        </tr>
+        </thead>
+        <tbody>
+        {% for index, file in collector.files %}
+            <tr class="file-info">
+                <td class="font-normal">
+                        <span class="icon">
+                            {% if file.type == 'pdf' %}
+                                {{ source('@SensiolabsGotenberg/Icon/file-type-pdf.svg') }}
+                            {% elseif file.type == 'screenshot' %}
```
Neirda24, line 100: ok. I thought maybe of using the file type itself. (png, jpeg, webp, pdf) and a fallback one ?

Neirda24 (commented):
Great work !

Neirda24:
Regarding `options` it can't be solved at the moment as there are no way to determine what are the options at the moment. the only thing keeping that state is an internal property `formFields`. Keep it as it is and will be updated in a follow-up pull request.

src/DataCollector/GotenbergDataCollector.php
ConstantBqt: Does it make sense to add it within the configuration array ? 
Screenshot from 2024-07-09 09-47-20

templates/Collector/sensiolabs_gotenberg.html.twig
Jean-Beru: I agree... For now, these values could be different when handling async or metadata retrieval.

src/DataCollector/GotenbergDataCollector.php
Jean-Beru: :+1: 
Since we can't retrieve `options`, I think that we can remove the corresponding line.

templates/Collector/sensiolabs_gotenberg.html.twig
Neirda24: even so. It will be a generated file except we won't have as much data.

Jean-Beru: Indeed, it can be a JSON file for metadata. So we can just focus on `requestCount` and use a label like `Request count".

src/DataCollector/GotenbergDataCollector.php
Jean-Beru: > yes options can be removed but that was not the point :)

Sure

> Regarding the builderId I would add it next to the Builder class. Maybe only if it is a custom one (not starting with a .

It will take something like 1/3 of the table width just for it. Not sure that this information is so useful :thinking:

Neirda24: because builders are services I think only showing classes may be confusing in some edge cases ? I agree that the class is the main intel in our case. so maybe hidden ? or a tooltip ?

ConstantBqt: A second line in the builder column with a hover that display the service Id ? 

The issue with this solution is that we can't copy the service Id. 

Screenshot from 2024-07-10 11-52-17

Neirda24: Indeed. Ok forget it. We'll see later.

Issue #43, opened 2024-05-28
Implement the read / write metadata routes


Neirda24:
https://gotenberg.dev/docs/routes#read-pdf-metadata-route + https://gotenberg.dev/docs/routes#write-pdf-metadata-route

PR #90, opened 2024-07-03
feat: improve response by using a stream

Jean-Beru:
### In a controller
```php
#[AsController]
final class PdfController
{
    #[Route(path: '/pdf', name: 'pdf', methods: ['GET'])]
    public function __invoke(
        GotenbergPdfInterface $gotenberg,
        Filesystem $filesystem,
        #[Autowire(env: 'resolve:PDF_DIR')] string $directory,
        LoggerInterface $logger,
    ): Response
    {
        return $gotenberg->html()
            ->header('pdf/header.html.twig')
            ->footer('pdf/footer.html.twig')
            ->content('pdf/invoice/pdf.html.twig', ['name' => 'Hubert'])
            ->landscape()
            ->fileName('my_pdf.pdf')
            ->processor(
                new ChainProcessor([
                    new FileProcessor($filesystem, $directory, $logger),
                    new DebugProcessor($logger),
                ]),
            )
            ->generate()
            ->stream()
        ;
    }
}
```

### In a command
```php
<?php

declare(strict_types=1);

namespace App\Command;

use Sensiolabs\GotenbergBundle\GotenbergPdfInterface;
use Sensiolabs\GotenbergBundle\Processor\FileProcessor;
use Symfony\Component\Console\Attribute\AsCommand;
use Symfony\Component\Console\Command\Command;
use Symfony\Component\Console\Input\InputInterface;
use Symfony\Component\Console\Output\OutputInterface;
use Symfony\Component\Console\Style\SymfonyStyle;
use Symfony\Component\DependencyInjection\Attribute\Autowire;
use Symfony\Component\Filesystem\Filesystem;

#[AsCommand(
    name: 'app:generate:invoice',
    description: 'Generates an invoice in PDF',
)]
class GenerateInvoiceCommand extends Command
{
    public function __construct(
        private readonly GotenbergPdfInterface $pdf,
        private readonly Filesystem $filesystem,
        #[Autowire(env: 'resolve:PDF_DIR')] private readonly string $directory,
    )
    {
        parent::__construct();
    }

    protected function execute(InputInterface $input, OutputInterface $output): int
    {
        $io = new SymfonyStyle($input, $output);

        // Intro
        $intro = $this->pdf->html()
            ->content('pdf/invoice/pdf.html.twig', ['name' => 'Hubert'])
            ->processor(new FileProcessor($this->filesystem, $this->directory))
            ->generate()
            ->process()
        ;
[...]

PR #93, opened 2024-07-08
Feature: Add a builder to write pdf metadata

ConstantBqt:
Add `WriteMetadataPdfBuilder`, inspired by `MergePdfBuilder`

Partially resolve issue #43

### DIFF
--- a/docs/pdf/write-metadata-builder.md
+++ b/docs/pdf/write-metadata-builder.md
@@ -1,6 +1,6 @@
-# Merge Builder
+# Write metadata Builder
 
-You may have the possibility to merge several PDF documents.
+You may have the possibility to write metadata within several PDF documents.
 
 ## Basic usage
 
@@ -18,7 +18,7 @@ You may have the possibility to merge several PDF documents.
     {
         public function yourControllerMethod(GotenbergPdfInterface $gotenberg): Response
         {
-            return $gotenberg->merge()
+            return $gotenberg->writeMetadata()
                 ->files(
                     'document.pdf',
                     'document_2.pdf',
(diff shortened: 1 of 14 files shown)


Neirda24 (commented):
I wonder though if it is worth the effort these routes ? Because we already provide a way to set metadata on generate. It requires to put the whole PDF into memory, send it though http to gotenberg and download it again as a stream response potentially. Might not be worth it. @Jean-Beru : WDYT ?

ConstantBqt:
> I wonder though if it is worth the effort these routes ? Because we already provide a way to set metadata on generate. It requires to put the whole PDF into memory, send it though http to gotenberg and download it again as a stream response potentially. Might not be worth it. @Jean-Beru : WDYT ?

IMO this is useful to update the metadata an existing PDF file. 
I imagine this scenario: 

> Someone needs to modify thousands of its company's PDFs to change the metadata Copyright or Keywords.

Neirda24:
> Someone needs to modify thousands of its company's PDFs to change the metadata Copyright or Keywords.

But then I would recommend to directly use a binary for this scenario. It requires too much memory + http bandwidth for such small and edge cases...

Jean-Beru:
> I wonder though if it is worth the effort these routes ? Because we already provide a way to set metadata on generate. It requires to put the whole PDF into memory, send it though http to gotenberg and download it again as a stream response potentially. Might not be worth it. @Jean-Beru : WDYT ?

Putting whole PDF in memory is not mandatory. HttpClient accepts callback and resources as body: https://symfony.com/doc/current/http_client.html#uploading-data.

 > But then I would recommend to directly use a binary for this scenario. It requires too much memory + http bandwidth for such small and edge cases...

Of course but some applications may want to delegate this task to the Gotenberg container.

IMHO ([here](https://www.urbandictionary.com/define.php?term=IMHO) @ConstantBqt :trollface: ), this endpoint can be implemented as a very simple builder. The hardest thing is to know how split our code between file generation (screenshot, pdf, etc.) and data retrieval (read metadata, async, etc.)

Neirda24:
Our implementation of the http client does not yet covers memory efficient input. even with https://github.com/sensiolabs/GotenbergBundle/pull/90 we will only cover the output part.

Jean-Beru:
> Our implementation of the http client does not yet covers memory efficient input. even with #90 we will only cover the output part.

Sure! It deserve a new issue to improve content and asset files.

docs/pdf/write-metadata-builder.md
```diff
@@ -0,0 +1,91 @@
+# Write metadata Builder
```
Neirda24, line 1: could you also update the `docs/generate.php`, run it and commit the changes please ? Thank  you (cc @Jean-Beru we probably need a `CONTRIBUTING` file

docs/pdf/write-metadata-builder.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Basic usage
+
+> [!WARNING]  
+> As assets files, by default the PDF files are fetch in the assets folder of
+> your application.  
+> For more information about path resolution go to [assets documentation](../assets.md).
+
+```php
+    namespace App\Controller;
```
Neirda24, line 13: we are updating docs to remove those indents.

docs/pdf/write-metadata-builder.md
```diff
@@ -0,0 +1,91 @@
...
+
+    use Sensiolabs\GotenbergBundle\GotenbergPdfInterface;
+
+    class YourController
+    {
+        public function yourControllerMethod(GotenbergPdfInterface $gotenberg): Response
+        {
+            return $gotenberg->writeMetadata()
```
Neirda24, line 21: looking at the example here I wonder. Because there is also the read metadata. We don't "generate" anything. Maybe it should out of scope of builders... ?

docs/pdf/write-metadata-builder.md
Neirda24: https://github.com/sensiolabs/GotenbergBundle/issues/95

Neirda24:
closing because must be redone using newer structure

Neirda24:
thank you @ConstantBqt

PR #97, opened 2024-07-12
[RELEASE_8.8.0] Add new fileds for LibreOfficeBuilder

StevenRenaux:
Related to :
https://github.com/gotenberg/gotenberg/releases/tag/v8.8.0
https://github.com/gotenberg/gotenberg/releases/tag/v8.7.0


Jean-Beru (approved):

src/Builder/Pdf/LibreOfficePdfBuilder.php
```diff
@@ -147,14 +147,184 @@ public function metadata(array $metadata): static
...
+    {
+        $this->formFields['allowDuplicateFieldNames'] = $bool;
+
+        return $this;
+    }
+
+    /**
+     * Specify if bookmarks are exported to PDF.
+     */
+    public function exportBookmarks(bool $bool = false): self
```
Neirda24, line 171: so it's `true` by default ? Because the way we designed those boolean methods is usually to have as a default argument the inverse of the default from gotenberg

src/Builder/Pdf/LibreOfficePdfBuilder.php
```diff
@@ -147,14 +147,184 @@ public function metadata(array $metadata): static
...
+     */
+    public function reduceImageResolution(bool $bool = true): self
+    {
+        $this->formFields['reduceImageResolution'] = $bool;
+
+        return $this;
+    }
+
+    /**
+     * If the form field reduceImageResolution is set to true, tell if all images will be reduced to the given value in DPI. Possible values are: 75, 150, 300, 600 and 1200.
```
Neirda24, line 319: because there is a fixed list. maybe either a constant list (to allow for further numbers), an enum|int (allowing both a fixed value + something else)

src/Builder/Pdf/LibreOfficePdfBuilder.php
```diff
@@ -147,14 +147,184 @@ public function metadata(array $metadata): static
...
+    {
+        $this->formFields['losslessImageCompression'] = $bool;
+
+        return $this;
+    }
+
+    /**
+     * Specify the quality of the JPG export. A higher value produces a higher-quality image and a larger file. Between 1 and 100.
+     */
+    public function quality(int $quality): self
```
Neirda24, line 301: `@param int<1, 100> $quality`

src/Builder/Pdf/LibreOfficePdfBuilder.php
StevenRenaux: yes, like exportFormFields too
https://gotenberg.dev/docs/routes#page-properties-libreoffice

Neirda24: maybe `doNotExportFormFields` then ? WDYT ?

StevenRenaux: You're right it's more logical, I just thought to follow the same naming but it's not that clear for a `false` value.

StevenRenaux:
@Neirda24 Done :wink:

src/Builder/Pdf/LibreOfficePdfBuilder.php
StevenRenaux: Check added in `configuration.php` also

src/Builder/Pdf/LibreOfficePdfBuilder.php
```diff
@@ -64,7 +65,7 @@ public function nativePageRanges(string $range): self
     /**
      * Set whether to export the form fields or to use the inputted/selected content of the fields.
      */
-    public function exportFormFields(bool $bool = false): self
+    public function doNotExportFormFields(bool $bool = false): self
```
Neirda24, line 68: maybe worth having both ? `doNotExportFormFields()` and `exportFormFields()`

Neirda24: let's keep it like it is now.

Neirda24 (approved):

PR #98, opened 2024-07-12
[RELEASE_8.7.0] Add new  field form Chromium

StevenRenaux:
Related to:
https://github.com/gotenberg/gotenberg/releases/tag/v8.7.0


Jean-Beru (approved):

docs/pdf/customization.md
```diff
@@ -707,6 +708,30 @@ class YourController
...
+
+use Sensiolabs\GotenbergBundle\GotenbergScreenshotInterface;
+
+class YourController
+{
+    public function yourControllerMethod(GotenbergScreenshotInterface $gotenberg): Response
+    {
+        return $gotenberg
+            ->html()
+            ->userAgent('Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko)')
```
Neirda24, line 728: it would be great if we could also have constants with some basic userAgent. Maybe we should have an object that represent Devices. Like iPhone, etc etc. It would set the width, height, user agent, and so on. WDYT @StevenRenaux , @Jean-Beru ?

For this PR maybe simply the constants from https://deviceatlas.com/blog/list-of-user-agent-strings or similar ?

StevenRenaux: I was thinking of adding it into a trait, but in 8.1 we can't add a constant in a trait.

Do you think this is relevant because it would make a long constant list to add in two `abstractChromium` at the moment?

Neirda24: No I would create 

```
enum UserAgent
{
    public const IPHONE_4S = 'xxx';
}
```

in another directory than builder (enum namespace maybe ?) beware that we use `enum` php struct to avoid being able to extend / overrid constants but we do use the enum as a constant list only.

StevenRenaux: Ok, I didn't understand that way

StevenRenaux: I also let the possibility to pass a string if the enum doesn't fit the need

PR #99, opened 2024-07-14
Update README

StevenRenaux:
Update with  symfony flex info.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,17 +21,25 @@ a screenshot.
 
 ## How to install
 
-> [!CAUTION]
-> To use this bundle, you first need to install and configure [Gotenberg 8.x](https://gotenberg.dev/docs/getting-started/installation).
-
 Install the bundle using composer :
 
 ```bash
 composer require sensiolabs/gotenberg-bundle
 ```
 
-If not using Symfony Flex, enable the bundle by adding it to the list of
-registered bundles in the ``config/bundles.php`` file of your project:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,17 +21,25 @@ a screenshot.
...
 composer require sensiolabs/gotenberg-bundle
 ```
 
-If not using Symfony Flex, enable the bundle by adding it to the list of
-registered bundles in the ``config/bundles.php`` file of your project:
+### With Symfony Flex
+
+If you accept the Symfony Flex recipe during installation, the bundle is registered, the configuration
+skeleton file is created, the .env file is updated with `GOTENBERG_DSN` and dockerfile is also created
+to get a gotenberg image, you need to configure the ports related to `GOTENBERG_DSN` in your .env file.
```
Neirda24, line 34:
```suggestion
to get the [gotenberg image](https://hub.docker.com/r/gotenberg/gotenberg), you need to configure the ports related to `GOTENBERG_DSN` in your .env file. 
```

Neirda24 (approved):

README.md
Jean-Beru, line 34: The host could be updated too since it's `localhost` by default. If your script is ran in a container, the host will the `gotenberg`.

We could also mention that the DSN can be set using an environment variable.

Jean-Beru (approved):

PR #103, opened 2024-07-19
Fix monthly downloads badge

smnandre:


Jean-Beru (approved):

Jean-Beru:
Good catch. Thanks @smnandre !

PR #106, opened 2024-09-06
fix: do not added inches unit for Gotenberg compatbility (<8.3)

pierregaste:


pierregaste:
Gotenberg added the notion of unit for margins and page sizes starting with version 8.3 (https://github.com/gotenberg/gotenberg/releases/tag/v8.3.0).

On older versions, if you specify margins and/or a page size, you get an error from Gotenberg:

> Invalid form data: form field 'marginTop' is invalid (got '0mm', resulting to strconv.ParseFloat: parsing "0mm": invalid syntax)

For these older versions, you can only work with inches. I therefore suggest not sending the "inch" unit in order to maintain backward compatibility on these versions.

In my case, it is Platform.sh that is still stuck in version 8.2.

Neirda24:
Hi @pierregaste . Thank you for your contribution.

we discussed this internally. Supporting previous version is not only about sending units or not. Some methods would have to not be usable or with a warning. this is too much complexity to handle right now.

regarding the few lines that were modified. we could indeed make it optional to not send something that will default natively on gotenberg side. then this would have to be applied to other methods as well. could you investigate a bit which should be changed ? I'll try to find some time late next week to investigate myself.

pierregaste:
> Hi @pierregaste . Thank you for your contribution.
> 
> we discussed this internally. Supporting previous version is not only about sending units or not. Some methods would have to not be usable or with a warning. this is too much complexity to handle right now.
> 
> regarding the few lines that were modified. we could indeed make it optional to not send something that will default natively on gotenberg side. then this would have to be applied to other methods as well. could you investigate a bit which should be changed ? I'll try to find some time late next week to investigate myself.

Yes, I agree! Make most sense in this way. I'll try to add this option in this PR instead of this previous modifications.

Neirda24:
Hi @pierregaste . We made changes here https://github.com/sensiolabs/GotenbergBundle/pull/173 thanks to the rework on 1.0.0. It allowed us to configure based on gotenberg version. Let us know if its working as you wish. Closing this for the moment. Feel free to open an issue if new problem were found.

PR #112, opened 2024-10-05
Add composer analyser checker

Neirda24:


Neirda24:
PHPStan issue is unrelated. Fixed in another PR.

Jean-Beru (approved):
Great, I did not know this tool👌

.php-cs-fixer.dist.php
```diff
@@ -6,7 +6,9 @@
         __DIR__ . '/docs',
         __DIR__ . '/src',
         __DIR__ . '/tests',
+        __DIR__ . '/',
```
Jean-Beru, line 9: Since the Finder is recursive, defining previous paths is useless

Neirda24: It was to include php files at the root of the library. I hesitated. Should we include php files at the root of the library in cs fixer too ?

PR #118, opened 2024-11-16
Poc/builder v2

Jean-Beru:
WIP with limited feature and a lot of old/useless files

- [ ] Create all traits
- [ ] Create all builder using traits
- [ ] Auto guess builder semantic config
- [ ] Auto guess semantic nodes (using Type component ? => ->info; ->examples; type)
- [ ] Update all tests (create one test trait per trait)
- [ ] Fix configurator for builders (use variadic ?)
- [ ] Fix PHPStan
- [ ] Doc on how to create your own builder / config
- [ ] Fix profiler
- [ ] cleanup


config/builder.php
```diff
@@ -15,7 +16,13 @@
         ->abstract()
         ->args([
             service('sensiolabs_gotenberg.client'),
-            service('sensiolabs_gotenberg.asset.base_dir_formatter'),
+            service_locator([
+                'asset_base_dir_formatter' => service('sensiolabs_gotenberg.asset.base_dir_formatter'),
+                'logger' => service('logger')->nullOnInvalid(),
```
Neirda24, line 21: ⚠️ I think this will include the wrong logger service. Not the custom channel.

src/Builder/Behaviors/Chromium/AssetTrait.php
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Adds a file, like an image, font, stylesheet, and so on.
+     */
+    public function addAsset(string $path): static
+    {
+        $assets = $this->getBodyBag()->get('assets', []);
+
+        if (\array_key_exists($path, $assets)) {
+            return $this;
```
Neirda24, line 50: should we log a warning ? Also I think we should check if two files have the same name. I think gotenberg treat them as same file.

src/Builder/Behaviors/Chromium/CustomHttpHeadersTrait.php
```diff
@@ -0,0 +1,91 @@
...
+            return $this;
+        }
+
+        $current = $this->getBodyBag()->get('extraHttpHeaders', []);
+
+        $this->getBodyBag()->set('extraHttpHeaders', array_merge($current, $headers));
+
+        return $this;
+    }
+}
```
Neirda24, line 91: EOF

src/Builder/Behaviors/Chromium/CustomHttpHeadersTrait.php
```diff
@@ -0,0 +1,91 @@
...
+     */
+    public function addExtraHttpHeaders(array $headers): static
+    {
+        if ([] === $headers) {
+            return $this;
+        }
+
+        $current = $this->getBodyBag()->get('extraHttpHeaders', []);
+
+        $this->getBodyBag()->set('extraHttpHeaders', array_merge($current, $headers));
```
Neirda24, line 87: what about using an object like arrayiterator to always have a reference to push elements into ? Might be better than array_merge. Would also require normalizer to convert it to array

src/Builder/Behaviors/Chromium/EmulatedMediaTypeTrait.php
```diff
@@ -0,0 +1,36 @@
...
+            )
+        ;
+    }
+
+    /**
+     * Forces Chromium to emulate, either "screen" or "print". (default "print").
+     */
+    public function emulatedMediaType(EmulatedMediaType $mediaType): static
+    {
+        $this->getBodyBag()->set('emulatedMediaType', $mediaType);
```
Neirda24, line 32: `->value` is missing or am I missing something ? (allowedValues)

+ EOF

src/Builder/Behaviors/Dependencies/AssetBaseDirFormatterAwareTrait.php
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Sensiolabs\GotenbergBundle\Builder\Behaviors\Dependencies;
+
+use Sensiolabs\GotenbergBundle\Formatter\AssetBaseDirFormatter;
+
+trait AssetBaseDirFormatterAwareTrait
+{
+    use DependencyAwareTrait;
+
+    protected function getAssetBaseDirFormatter(): AssetBaseDirFormatter
```
Neirda24, line 11: should maybe directly be part of DependencyAwareTrait. WDYT ?

src/Builder/Behaviors/Dependencies/LoggerAwareTrait.php
```diff
@@ -0,0 +1,20 @@
...
+use Psr\Log\LoggerInterface;
+use Twig\Environment;
+
+trait LoggerAwareTrait
+{
+    use DependencyAwareTrait;
+
+    protected function getLogger(): LoggerInterface|null
+    {
+        if (($logger = $this->dependencies->get('logger')) instanceof Environment) {
```
Neirda24, line 14:
```suggestion
        if (($logger = $this->dependencies->get('logger')) instanceof LoggerInterface) {
```

src/Builder/Behaviors/Dependencies/LoggerAwareTrait.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Sensiolabs\GotenbergBundle\Builder\Behaviors\Dependencies;
+
+use Psr\Log\LoggerInterface;
+use Twig\Environment;
```
Neirda24, line 6:
```suggestion
```

tests/Builder/MergePdfBuilderTest.php
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * @extends GotenbergBuilderTestCase<MergePdfBuilder>
+ */
+#[CoversClass(MergePdfBuilder::class)]
+class MergePdfBuilderTest extends GotenbergBuilderTestCase
+{
+    protected function createBuilder(GotenbergClientInterface $client, ContainerInterface $dependencies): BuilderInterface
+    {
+        $dependencies->set('asset_base_dir_formatter', new AssetBaseDirFormatter(new Filesystem(), __DIR__, 'fixtures'));
```
Jean-Beru, line 22: Could be mocked if `AssetBaseDirFormatter` is reworked (final and without interface for now)

src/Builder/AbstractBuilder.php
```diff
@@ -0,0 +1,118 @@
...
+        $this->processor = $processor;
+
+        return $this;
+    }
+
+    public function generate(): GotenbergFileResult
+    {
+        $response = $this->client->call($this->getEndpoint(), $this->bodyBag, $this->headersBag);
+
+        if (!\in_array($response->getStatusCode(), [200, 204], true)) {
```
Neirda24, line 72: hmmm I don't see the benefit of having this here instead of in the call method. I was thinking more in the GileResult object. WDYT ?

Jean-Beru: Done

src/Builder/Behaviors/Chromium/AssetTrait.php
Jean-Beru: We could, but I don't think that it will create useless logs when your Twig template use `gotenberg_asset` multiple times. For example, a logo displayed in the first and the last page.

Neirda24: you are right. maybe if both the filename is the same but the path is different

Jean-Beru: AFAIK, we can only use files under the root folder. It could be possible but we'll have to generate a unique ID depending on the file path when adding (and using) the asset.

src/Builder/Behaviors/Chromium/CustomHttpHeadersTrait.php
Jean-Beru: Good idea. Would it be possible to replace an existing header with it?

src/Builder/Behaviors/Dependencies/AssetBaseDirFormatterAwareTrait.php
Jean-Beru: Will this dependency always be required ? Maybe the read metadata API will be the only API with no `files` form data 🤔

src/Builder/Behaviors/Chromium/CustomHttpHeadersTrait.php
Neirda24: Normally yes. I don't see why not.

src/Builder/Behaviors/Dependencies/AssetBaseDirFormatterAwareTrait.php
Neirda24: Clearly not. But I think it does not clutter too much the builder files and simplify things a lot.

src/Builder/Behaviors/Chromium/AssetTrait.php
Neirda24: This is somethign we need to discuss in further improvements : if two files with same name but different path are added we should either handle it with dynamic filename hash or throw an eroor / warning.

Neirda24:
this can be closed right @Jean-Beru ?

Jean-Beru:
I think so

PR #119, opened 2024-11-20
[RELEASE_8.14.0] Add new field `generateDocumentOutline`

StevenRenaux:


Neirda24 (approved):

src/DependencyInjection/Configuration.php
```diff
@@ -229,6 +229,10 @@ private function addChromiumPdfOptionsNode(ArrayNodeDefinition $parent): void
                     ->info('Define whether to prefer page size as defined by CSS - default false. https://gotenberg.dev/docs/routes#page-properties-chromium')
                     ->defaultNull()
                 ->end()
+                ->booleanNode('generate_document_outline')
+                    ->info('Define whether the document outline should be embedded into the PDF - default false. https://gotenberg.dev/docs/routes#page-properties-chromium')
+                    ->defaultNull()
```
Jean-Beru, line 234: Is it working with the default value since `generateDocumentOutline` does not accept `null` value ?

Neirda24: null and [] are filtered out in the Extension file. So it is not set to null. It is simply ignored by the configuration.

StevenRenaux: https://github.com/sensiolabs/GotenbergBundle/blob/bbdfaebcc6396a3a5ce1cd2c1ea11f497c63fdd1/src/DependencyInjection/SensiolabsGotenbergExtension.php#L132

Jean-Beru: :ok_hand:

Jean-Beru (approved):

PR #123, opened 2024-12-06
fix: doc generator builder case

rhyd42:
Case of 'Pdf' and 'Screenshot' causes broken links in generated documentation

### DIFF
--- a/docs/generate.php
+++ b/docs/generate.php
@@ -20,7 +20,7 @@ require_once \dirname(__DIR__).'/vendor/autoload.php';
  * @var array<string, non-empty-list<class-string>>
  */
 const BUILDERS = [
-    'Pdf' => [
+    'pdf' => [
         HtmlPdfBuilder::class,
         UrlPdfBuilder::class,
         MarkdownPdfBuilder::class,
@@ -28,7 +28,7 @@ const BUILDERS = [
         MergePdfBuilder::class,
         ConvertPdfBuilder::class,
     ],
-    'Screenshot' => [
+    'screenshot' => [
         HtmlScreenshotBuilder::class,
         UrlScreenshotBuilder::class,
         MarkdownScreenshotBuilder::class,


docs/generate.php
```diff
@@ -20,15 +20,15 @@
  * @var array<string, non-empty-list<class-string>>
  */
 const BUILDERS = [
-    'Pdf' => [
+    'pdf' => [
```
Jean-Beru, line 23: `Pdf` should be kept since it's used to generate the title.

`pdf` and `screenshot` folders should be renamed to fix links. Also, the CI check should use `git status --short --renames` instead to check the diff. `git diff --name-only` does not seem to detect file renaming.

rhyd42: If renaming pdf and screenshot, should the async folder also be renamed to Async for consistency?

Jean-Beru: Good catch! We can also fix this folder.

Or, in order to keep folders in lowercase, the `BUILDERS` constant could keep lowercase keys but using `ucfirst` or `ucwords` when [writing the title](https://github.com/sensiolabs/GotenbergBundle/blob/ee3376655cfdbb3991fb0862ead13a9078e4645d/docs/generate.php#L150). WDYT ?

rhyd42: `ucfirst` or `ucwords `was my first thought, but looking at the other folders there's a mix of snake case and Pascal case and lowercase, and even the folder hierarchy is different for /docs/. Maybe a new issue should be opened to decide on consistency for cases of file / folder names.
I was hoping to suggest a quick fix for the broken links but I seem to have opened a can of worms!

/docs/pdf/builders_api/HtmlPdfBuilder.md - _lower case, snake case_
/src/Builder/Pdf/HtmlPdfBuilder.php - _Pascal case_
/tests/Builder/Pdf/HtmlPdfBuilderTest.php - _Pascal case_
/test/Fixtures/pdf/ - _Pascal case, lower case_

Jean-Beru: IMO, /docs/pdf/builders_api/HtmlPdfBuilder.md  should be the correct way since only `HtmlPdfBuilder` refers to a classname. So, the `ucfrist` solution is the best for now.

Concerning `/test/Fixtures/pdf/`, you're probably right because this folder does not contains any class to load use PSR4. It could be done in a dedicated PR.

PR #124, opened 2024-12-12
Fix `header` and `Footer` configuration

StevenRenaux:


src/DependencyInjection/Configuration.php
```diff
@@ -173,11 +173,31 @@ private function addChromiumPdfOptionsNode(ArrayNodeDefinition $parent): void
                             ->info('Default header twig template to apply.')
                             ->defaultNull()
                         ->end()
-                        ->scalarNode('context')
+                        ->arrayNode('context')
                             ->info('Default context for header twig template.')
+                            ->useAttributeAsKey('name')
                             ->defaultValue([])
+                            ->arrayPrototype()
```
Neirda24, line 180: scalarPrototype is not enough ?

StevenRenaux: You can have multiple node (variables) into you context.
And with that I had the possibility to defined what we want. 
`name` of the variable
`value` of the variable

Or do I miss something?

Neirda24: Indeed my bad. Although I think having 

```yaml
context:
    key: some_value
```

is better than

```yaml
context:
    - {name: key, value: some_value}
```

so maybe : variablePrototype ?

StevenRenaux: @Neirda24 done ;)

src/DependencyInjection/Configuration.php
```diff
@@ -173,11 +173,20 @@ private function addChromiumPdfOptionsNode(ArrayNodeDefinition $parent): void
...
+                            ->variablePrototype()
+                            ->end()
                         ->end()
                     ->end()
+                    ->validate()
+                        ->ifTrue(static function ($array) {
+                            return null === $array['template'];
+                        })
+                        ->thenInvalid('A template must be provided.')
+                    ->end()
```
Jean-Beru, line 189: `->required()` on the `->scalarNode('template')` does not work?

Jean-Beru (approved):

src/DependencyInjection/Configuration.php
StevenRenaux: Didn't think about it and thought is was like cannotBeEmpty but finally not.

docs/configuration.md
```diff
@@ -1201,6 +1201,30 @@ sensiolabs_gotenberg:
...
+            html:
+                header:
+                    template: 'header.html.twig'
+                    context:
+                        - { name: 'title', value: 'Hello' }
+                        - { name: 'first_name', value: 'Jean Michel' }
+                footer:
+                    template: 'footer.html.twig'
+                    context:
+                        - { name: 'foo', value: 'bar' }
```
Jean-Beru, line 1225: If I remember well, we wanted to have this syntax, right?
```suggestion
                    context:
                        title: 'Hello'
                        first_name: 'Jean Michel'
                footer:
                    template: 'footer.html.twig'
                    context:
                        foo: 'bar'
```

src/DependencyInjection/Configuration.php
```diff
@@ -171,11 +171,14 @@ private function addChromiumPdfOptionsNode(ArrayNodeDefinition $parent): void
                     ->children()
                         ->scalarNode('template')
                             ->info('Default header twig template to apply.')
-                            ->defaultNull()
+                            ->isRequired()
+                            ->cannotBeEmpty()
                         ->end()
-                        ->scalarNode('context')
+                        ->arrayNode('context')
```
Neirda24, line 177: Might be worth using https://symfony.com/doc/current/components/config/definition.html#array-node-options `normalizeKeys(false)`

src/DependencyInjection/Configuration.php
```diff
@@ -184,11 +187,14 @@ private function addChromiumPdfOptionsNode(ArrayNodeDefinition $parent): void
                     ->children()
                         ->scalarNode('template')
                             ->info('Default footer twig template to apply.')
-                            ->defaultNull()
+                            ->isRequired()
```
Neirda24, line 190: required ? Meaning the developper MUST provide a value.

src/DependencyInjection/Configuration.php
StevenRenaux: As it was done in https://github.com/sensiolabs/GotenbergBundle/pull/130/files#diff-ebe00144d6cecb5ca8a2860883e01147f42bbed2b12a3a7f44915908a4829faaR813 then.
 I thought to force the best practice of the variable naming, but you right it's not our role to do that.

src/DependencyInjection/Configuration.php
StevenRenaux: if `header` of `footer` is not displayed into the user config it's not required because the node is not added.
If the user add `header` for example it become required.

Neirda24: ok great then !

src/DependencyInjection/Configuration.php
StevenRenaux: @Neirda24 done!

Neirda24 (approved):

PR #125, opened 2024-12-12
Handle assets from the Symfony component

### LINKED ISSUE #121: An exception has been thrown during the rendering of a template ("Warning: Undefined array key "_builder"").
When I try to use the {{ gotenberg_asset('public/...') }} I get this error

`An exception has been thrown during the rendering of a template ("Warning: Undefined array key "_builder"").`

I have no idea what to do. I just installed the bundle, it works with url, i tried with html, works without assets, now breaks when loading assets.
Any help ?
Symfony 6.4.15, and latest gotenberg bundle

Jean-Beru:
Resolves #121 

Allow using the `asset` twig function from the TwigBridge for public assets.

### DIFF
--- a/src/Twig/AssetExtension.php
+++ b/src/Twig/AssetExtension.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace Sensiolabs\GotenbergBundle\Twig;
+
+use Sensiolabs\GotenbergBundle\Builder\Pdf\AbstractChromiumPdfBuilder;
+use Sensiolabs\GotenbergBundle\Builder\Screenshot\AbstractChromiumScreenshotBuilder;
+use Sensiolabs\GotenbergBundle\Exception\RenderingException;
+use Symfony\Component\Asset\Packages;
+use Symfony\Bridge\Twig\Extension\AssetExtension as TwigAssetExtension;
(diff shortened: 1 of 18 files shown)


src/Twig/AssetExtension.php
```diff
@@ -0,0 +1,63 @@
...
+    ) {
+    }
+
+    public function getFunctions(): array
+    {
+        $functions = [];
+
+        foreach ($this->inner->getFunctions() as $function) {
+            if ('asset' === $function->getName()) {
+                $function = new TwigFunction('asset', $this->getAsset(...), ['needs_context' => true]);
```
smnandre, line 29: So this extension will decorate my entire app usage of "asset" ? ... well...

Jean-Beru: Since the `PathPackage` does not include the host, `asset` cannot work in a Gotenberg context.

How can we make it working without decorating it ? I was also thinking about the `<base>` tag to solve this problem but [it's not recommended](https://gotenberg.dev/docs/routes#html-file-into-pdf-route).

Jean-Beru:
> A custom asset package would not work here ?

If this package uses `base_path`, Gotenberg won't be able to find the file without the domain.

PR #127, opened 2024-12-22
Use `Path:: methods` instead of Filesystem instance

smnandre:
`Filesystem->isAbsolutePath` does not check if the file exists and just make quick checks on the string.

So here the Path utility methods seem a better choice as they are platform-agnostic and allow safe operations like `::join`  handle `../../` in paths and handle full path canonicalization.

And the tests can be done on any strings, independently on the Fixtures files here.

I think this method is used before calling `new File()`, so maybe the whole thing could be replaced by some factory/provider returning an Asset or something like that?

I've not touched the constructor (BC) or the DI for now, depending on your feedback & vision on this.

### DIFF
--- a/src/Formatter/AssetBaseDirFormatter.php
+++ b/src/Formatter/AssetBaseDirFormatter.php
@@ -3,6 +3,7 @@
 namespace Sensiolabs\GotenbergBundle\Formatter;
 
 use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Filesystem\Path;
 
 final class AssetBaseDirFormatter
 {
@@ -18,14 +19,14 @@ final class AssetBaseDirFormatter
 
     public function resolve(string $path): string
     {
-        if ($this->filesystem->isAbsolutePath($path)) {
+        if (Path::isAbsolute($path)) {
             return $path;
         }
 
-        if ($this->filesystem->isAbsolutePath($this->baseDir)) {
-            return "{$this->baseDir}/{$path}";
+        if (Path::isAbsolute($this->baseDir)) {
+            return Path::join($this->baseDir, $path);
         }
 
-        return "{$this->projectDir}/{$this->baseDir}/{$path}";
+        return Path::join($this->projectDir, $this->baseDir, $path);
     }
 }
--- a/src/Formatter/AssetBaseDirFormatter.php
+++ b/src/Formatter/AssetBaseDirFormatter.php
@@ -2,15 +2,16 @@
 
 namespace Sensiolabs\GotenbergBundle\Formatter;
 
-use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Filesystem\Path;
 
+/**
+ * @internal
+ */
 final class AssetBaseDirFormatter
 {
     private readonly string $baseDir;
 
     public function __construct(
-        private readonly Filesystem $filesystem,
         private readonly string $projectDir,
         string $baseDir,
     ) {
--- a/tests/Formatter/AssetBaseDirFormatterTest.php
+++ b/tests/Formatter/AssetBaseDirFormatterTest.php
@@ -12,8 +12,8 @@ use Symfony\Component\Filesystem\Filesystem;
 #[CoversClass(AssetBaseDirFormatter::class)]
 final class AssetBaseDirFormatterTest extends TestCase
 {
-    private const BASE_PATH = "/var/www/project";
-    
+    private const BASE_PATH = '/var/www/project';
+
     /**
      * @return iterable<string, array<int, string>>
      */
--- a/tests/Formatter/AssetBaseDirFormatterTest.php
+++ b/tests/Formatter/AssetBaseDirFormatterTest.php
@@ -12,16 +12,18 @@ use Symfony\Component\Filesystem\Filesystem;
 #[CoversClass(AssetBaseDirFormatter::class)]
 final class AssetBaseDirFormatterTest extends TestCase
 {
+    private const BASE_PATH = "/var/www/project";
+    
     /**
      * @return iterable<string, array<int, string>>
      */
     public static function generateBaseDirectoryAndPath(): iterable
     {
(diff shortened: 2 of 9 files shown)


Jean-Beru:
> I've not touched the constructor (BC) or the DI for now, depending on your feedback & vision on this.

It's a BC break but the only way to be affected is to redefine the 1st argument in a CompilerPass. I'm not sure that someone already did that. And this bundle is experimental :angel: .

BTW, this service should be marked as internal to encourage decoration.

For me, we can update this service but we can wait for other maintainers opinion.

smnandre:
I think this package should... not be kept like this. 

__Right now__ it's not much more than a static helper.

src/Formatter/AssetBaseDirFormatter.php
```diff
@@ -3,6 +3,7 @@
 namespace Sensiolabs\GotenbergBundle\Formatter;
 
 use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Filesystem\Path;
 
 final class AssetBaseDirFormatter
```
Neirda24, line 8: as it is final without any interface, we should make this internal for the moment. When we will be working on Flysystem integration this class will change.

tests/Formatter/AssetBaseDirFormatterTest.php
```diff
@@ -12,24 +12,26 @@
...
+        yield 'relative path and relative base dir with end slash' => ['document.odt', 'bar/baz/', self::BASE_PATH.'/bar/baz/document.odt'];
     }
 
     #[DataProvider('generateBaseDirectoryAndPath')]
     #[TestDox('Resolve path when "$_dataName"')]
     public function testResolvePathCorrectly(string $path, string $baseDirectory, string $expectedResult): void
     {
         $filesystem = new Filesystem();
-        $assetBaseDirFormatter = new AssetBaseDirFormatter($filesystem, __DIR__.'/../Fixtures', $baseDirectory);
+        $assetBaseDirFormatter = new AssetBaseDirFormatter($filesystem, self::BASE_PATH, $baseDirectory);
```
Neirda24, line 34: IMO : `filesystem` could be safely removed

src/Formatter/AssetBaseDirFormatter.php
```diff
@@ -2,15 +2,16 @@
 
 namespace Sensiolabs\GotenbergBundle\Formatter;
 
-use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Filesystem\Path;
 
+/**
+ * @internal
```
Neirda24, line 8: thanks. If the class is internal might be worth also prefixing the service with `.` to make it "hidden" WDYT ?

smnandre: Done

tests/Formatter/AssetBaseDirFormatterTest.php
```diff
@@ -7,29 +7,29 @@
...
      * @return iterable<string, array<int, string>>
      */
     public static function generateBaseDirectoryAndPath(): iterable
     {
-        yield 'absolute path and absolute base dir' => [__DIR__.'/../Fixtures/assets/file.md', __DIR__.'/../Fixtures/assets', __DIR__.'/../Fixtures/assets/file.md'];
-        yield 'absolute path and relative base dir' => [__DIR__.'/../Fixtures/assets/logo.png', '/assets', __DIR__.'/../Fixtures/assets/logo.png'];
-        yield 'relative path and relative base dir' => ['document.odt', 'assets/office', __DIR__.'/../Fixtures/assets/office/document.odt'];
-        yield 'relative path and relative base dir with end slash' => ['document.odt', 'assets/office/', __DIR__.'/../Fixtures/assets/office/document.odt'];
-        yield 'relative path and absolute base dir' => ['office/document_1.docx', __DIR__.'/../Fixtures/assets', __DIR__.'/../Fixtures/assets/office/document_1.docx'];
+        yield 'absolute path and absolute base dir' => [self::BASE_PATH.'/foo/file.md', self::BASE_PATH.'/bar', self::BASE_PATH.'/foo/file.md'];
```
Jean-Beru, line 21: Even if the base path is reused in some data, it could be easier to read to avoid concatenation.

```php
yield 'absolute path and absolute base dir' => ['/srv/foo/file.md', '/bar', '/srv/foo/file.md'];
```

WDYT ?

tests/Formatter/AssetBaseDirFormatterTest.php
```diff
@@ -7,29 +7,29 @@
 use PHPUnit\Framework\Attributes\TestDox;
 use PHPUnit\Framework\TestCase;
 use Sensiolabs\GotenbergBundle\Formatter\AssetBaseDirFormatter;
-use Symfony\Component\Filesystem\Filesystem;
 
 #[CoversClass(AssetBaseDirFormatter::class)]
 final class AssetBaseDirFormatterTest extends TestCase
 {
+    private const BASE_PATH = '/var/www/project';
```
Neirda24, line 14: seems very inherent of your install.

tests/Formatter/AssetBaseDirFormatterTest.php
Neirda24: the issue is : we need to fetch some files but also to make sure some files doesn't exists. not sure the changes from `__DIR__` is better. Why not keep `__DIR__` ?

tests/Formatter/AssetBaseDirFormatterTest.php
smnandre: ? It's fixtures :)

smnandre:
> the issue is : we need to fetch some files but also to make sure some files doesn't exists. not sure the changes from __DIR__ is better. Why not keep __DIR__ ?

I don't think this is the responsability of this class. 

> seems very inherent of your install.

The opposite actually. But this proves the role and effect of this class were confusing.

As I said in the original message, previous implementation _did not_ check file existence.

Neirda24 (approved):

Neirda24:
thanks @smnandre ❤️

PR #128, opened 2024-12-22
Add `GotenbergAssetRuntime` to avoid passing Builder in context

smnandre:
Passing the builder in the Twig context is a bit fragile. 

It won't work when template uses any of Twig layout features (includes, embed, macros, ...) without passing the full context.

Also, starting with Twig 4.0, the `include` function won't pass the full context per default, with no "boolean" to do so automatically, encouraging template isolation and composition.

This PR introduce a Twig Runtime, lowering the impact of the Extension on the global scope, and allowing to create a time-limited "scope", during which the `gotenberg_asset` will work as expected.

### DIFF
--- a/src/Twig/GotenbergAssetRuntime.php
+++ b/src/Twig/GotenbergAssetRuntime.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace Sensiolabs\GotenbergBundle\Twig;
+
+use Sensiolabs\GotenbergBundle\Builder\Pdf\AbstractChromiumPdfBuilder;
+use Sensiolabs\GotenbergBundle\Builder\Screenshot\AbstractChromiumScreenshotBuilder;
+
+/**
+ * @internal
+ */
+final class GotenbergAssetRuntime
+{
+    private AbstractChromiumPdfBuilder|AbstractChromiumScreenshotBuilder|null $builder = null;
+
+    public function setBuilder(AbstractChromiumPdfBuilder|AbstractChromiumScreenshotBuilder|null $builder): void
+    {
+        $this->builder = $builder;
+    }
+
+    /**
+     * This function is used to get the URL of an asset during the rendering
+     * of a PDF or a screenshot with the Gotenberg client.
+     *
+     * It only works if the builder is an instance of AbstractChromiumPdfBuilder
+     * or AbstractChromiumScreenshotBuilder.
+     */
+    public function getAssetUrl(string $path): string
+    {
+        if (null === $this->builder) {
+            throw new \LogicException('You need to extend from AbstractChromiumPdfBuilder to use "gotenberg_asset" function.');
+        }
+
+        $this->builder->addAsset($path);
+
+        return basename($path);
+    }
+}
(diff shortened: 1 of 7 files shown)


src/Twig/GotenbergAssetRuntime.php
```diff
@@ -0,0 +1,37 @@
...
+     * This function is used to get the URL of an asset during the rendering
+     * of a PDF or a screenshot with the Gotenberg client.
+     *
+     * It only works if the builder is an instance of AbstractChromiumPdfBuilder
+     * or AbstractChromiumScreenshotBuilder.
+     */
+    public function getAssetUrl(string $path): string
+    {
+        if (null === $this->builder) {
+            throw new \LogicException('You need to extend from AbstractChromiumPdfBuilder to use "gotenberg_asset" function.');
```
Jean-Beru, line 30:
```suggestion
            throw new \LogicException('The gotenberg_asset function must be used in a Gotenberg context. You can use it via the AbstractChromiumPdfBuilder or AbstractChromiumScreenshotBuilder class.');
```
Should be easier to understand.

smnandre: I took the original one, but yeah this looks a bit weird. 

To be honest i would not specify class names here at all.. It's a bit weird no ?  

And add a dedicated interface/VO like "Gotenberg\Assets" (or "AssetSet" or "TemplateAssetBag" or ...)  and just use $assets->add(...).  

Could even use a default one if none is set.

In the end, I think this function could very much work for everyone... the AssetBag would simply **not** be collected  after the render of the template. 

WDYT ?

Jean-Beru: > To be honest i would not specify class names here at all.. It's a bit weird no ?

Indeed it is not very helpful.

> And add a dedicated interface/VO like "Gotenberg\Assets" (or "AssetSet" or "TemplateAssetBag" or ...) and just use $assets->add(...).

I hope I can have some time to rework builders in that way. I started to work with a HeadersBag and a BodyBag, but an AssetBag should be added too.

smnandre: AssetBag->addAsset($asset)

Here, $asset can be:
* A logical path: mapped by assetmapper/webpack -> Package::getUrl.
* An absolute URL: directly accepted.
* A local file: either streamed or converted into a signed URL.

This approach should simplify things quite a bit.

And then a second twig function to inline _**some**_ assets (base64 images for instance, work good for header/footer things)

Neirda24 (approved):
Love it thanks @smnandre . Although it means that `gotenberg_asset` is clearly not existing within twig "normal" runtime. Which eventually we wanted to be able to enable that function to fallback to native "asset" if not in builder context. allowing the same twig template to be used for both web AND pdf. WDYT ?

smnandre:
I'm not immersed enough in this project to have many use cases in mind... but instinctively i would have said the need to have `asset` behave differently (only during Gotenberg render) is not that common.

Maybe best to have a distinct methods and user can extends their templates, blocks if they _do_ need to render all _but one image_ the same way.

🤷

PR #130, opened 2024-12-24
Update `extra_http_headers` configuration

StevenRenaux:
Use `variablePrototype`  for `extra_http_headers`.

We can use as usual

````yaml
pdf:
    html:
        extra_http_headers:
            - { name: 'My-Header', value: 'MyValue' }
````

Or also that syntax 

````yaml
pdf:
    html:
        extra_http_headers:
            My-Header: 'MyValue'
````

### CODE REVIEW

**Comment by Jean-Beru** [comment]
> BC break for old version

Maybe we can add a CHANGELOG.md to list new features or BC breaks. :thinking:

**Comment by smnandre** [comment]
I think you can have both syntax (with the useSomethingAsKey) no ?

**Review by Neirda24 — APPROVED** [review]

**Comment by Neirda24** [comment]
Approved but we need to investigate the BC break. Will try to spend some time on it during holidays

**Comment by smnandre** [comment]
cf example with "parameters" here https://symfony.com/doc/current/components/config/definition.html#appending-sections

**Comment by smnandre** [comment]
(worst case scenario, allow both syntaxes and transform them in beforeNormalization)

**Comment by Neirda24** [comment]
> (worst case scenario, allow both syntaxes and transform them in beforeNormalization)

Yes or in `Extension` class. Either way we need to decide if that BC break is important enough as we are in "experimental" still.

**Comment by StevenRenaux** [comment]
> I think you can have both syntax (with the useSomethingAsKey) no ?

@smnandre you right it's possible. An no BC break with that solution

**Review by Jean-Beru — APPROVED** [review]

File: docs/builders_api.md [hunk 1]
```diff
@@ -1,17 +1,17 @@
 # Builders API
 
-## Pdf
+## pdf
```
smnandre, line 3:
```suggestion
## Ppd
```

docs/builders_api.md
```diff
@@ -1,17 +1,17 @@
...
-* [ConvertPdfBuilder](./Pdf/builders_api/ConvertPdfBuilder.md)
+* [HtmlPdfBuilder](./pdf/builders_api/HtmlPdfBuilder.md)
+* [UrlPdfBuilder](./pdf/builders_api/UrlPdfBuilder.md)
+* [MarkdownPdfBuilder](./pdf/builders_api/MarkdownPdfBuilder.md)
+* [LibreOfficePdfBuilder](./pdf/builders_api/LibreOfficePdfBuilder.md)
+* [MergePdfBuilder](./pdf/builders_api/MergePdfBuilder.md)
+* [ConvertPdfBuilder](./pdf/builders_api/ConvertPdfBuilder.md)
 
-## Screenshot
+## screenshot
```
smnandre, line 12:
```suggestion
## Screenshot
```

docs/configuration.md
```diff
@@ -1205,6 +1135,17 @@ sensiolabs_gotenberg:
 
 HTTP headers to send by Chromium while loading the HTML document.
 
+```yaml
+sensiolabs_gotenberg:
+    default_options:
+        pdf:
+            html:
+                extra_http_headers:
+                    My-Header: 'MyValue'
```
smnandre, line 1144: We should avoid dash values here. Not even sure they work out of the box to be honest :|

```suggestion
                    'My-Header': 'MyValue'
```

docs/configuration.md
```diff
@@ -1214,14 +1155,14 @@ sensiolabs_gotenberg:
...
-Or Headers to send to your webhook endpoint
+Headers to send to your webhook endpoint
 
 ```yaml
 sensiolabs_gotenberg:
     webhook:
         default:
             extra_http_headers:
-                - { name: 'My-Header', value: 'MyValue' }
+                My-Header: 'MyValue'
```
smnandre, line 1165:
```suggestion
                'My-Header': 'MyValue'
```

docs/pdf/builders_api/HtmlPdfBuilder.md
```diff
@@ -50,6 +50,9 @@ Overrides the default margins (e.g., 0.39), in inches.
 * `preferCssPageSize(bool $bool)`:
 Define whether to prefer page size as defined by CSS. (Default false).
 
+* `generateDocumentOutline(bool $bool)`:
+Define whether the document outline should be embedded into the PDF. (Default false).
+
```
smnandre, line 55: Just to note: following method uses third person grammar style "prints" and not "print".

src/DependencyInjection/Configuration.php
```diff
@@ -736,33 +713,7 @@ private function addWebhookDeclarationNode(ArrayNodeDefinition $parent): void
...
-                                            ->ifTrue(static function ($option) {
-                                                return !\is_string($option);
-                                            })
-                                            ->thenInvalid('Invalid header value %s')
-                                        ->end()
-                                    ->end()
-                                ->end()
-                            ->end()
-                        ->end()
+                        ->append($this->addExtraHttpHeaders())
```
smnandre, line 716: Should probably be called addExtraHttpHeadersConfiguration or addExtraHttpHeadersConfigurationNodes no ?

tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -69,15 +69,15 @@ public function testWithExtraHeadersConfiguration(): void
                 'http_client' => 'http_client',
                 'default_options' => [
                     'pdf' => [
-                        'html' => ['extra_http_headers' => [['name' => 'MyHeader', 'value' => 'MyValue'], ['name' => 'User-Agent', 'value' => 'MyValue']]],
+                        'html' => ['extra_http_headers' => ['MyHeader' => 'MyValue', 'User-Agent' => 'MyAgent']],
```
smnandre, line 72: Should we keep the old format in test to ensure the compatibility really does work ?

docs/builders_api.md
Jean-Beru: Fixed in https://github.com/sensiolabs/GotenbergBundle/pull/133

docs/builders_api.md
Jean-Beru: Fixed in https://github.com/sensiolabs/GotenbergBundle/pull/133

docs/configuration.md
StevenRenaux: @smnandre it work with `normalizeKeys(false)` 
I didn't want to transform the value for after that re-transform it before sending the request. 

image

tests/DependencyInjection/ConfigurationTest.php
StevenRenaux: You right and it's done

docs/configuration.md
```diff
@@ -454,15 +444,7 @@ sensiolabs_gotenberg:
                         method:               null # One of "POST"; "PUT"; "PATCH"
 
                     # HTTP headers to send back to both success and error endpoints - default None. https://gotenberg.dev/docs/webhook
-                    extra_http_headers:
-
-                        # Example:
```
Neirda24, line 459: worth to keep examples as well for both new syntax and old syntax ?

tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -60,7 +60,24 @@ public function testInvalidNativePageRange(mixed $range): void
         );
     }
 
-    public function testWithExtraHeadersConfiguration(): void
+    /**
+     * @return iterable<string, array<array-key, array<string, string>|list<array<string, string>>>>
```
Neirda24, line 64:
```suggestion
     * @return iterable<string, list<array<array-key, string>|list<array{name: array-key, value: string}>>>
```

I think this should be good.

tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -60,7 +60,24 @@ public function testInvalidNativePageRange(mixed $range): void
...
+        yield 'with variable Prototype configuration' => [
+            ['MyHeader' => 'MyValue', 'User-Agent' => 'MyAgent'],
+        ];
+        yield 'with attribute as key configuration' => [
+            [['name' => 'MyHeader', 'value' => 'MyValue'], ['name' => 'User-Agent', 'value' => 'MyAgent']],
+        ];
+    }
+
+    /**
+     * @param array<array-key, array<string, string>|list<array<string, string>>> $configuration
```
Neirda24, line 77: same as previous

docs/configuration.md
StevenRenaux: done

smnandre (approved):
👍

PR #132, opened 2025-01-03
[RELEASE_8.15.0] Add split feature

StevenRenaux:
Related to [RELEASE_8.15.0](https://github.com/gotenberg/gotenberg/releases/tag/v8.15.0)


docs/Pdf/builders_api/HtmlPdfBuilder.md
```diff
@@ -0,0 +1,168 @@
+# HtmlPdfBuilder
```
Neirda24, line 1: problem while generating I think.

docs/configuration.md
```diff
@@ -231,6 +231,15 @@ sensiolabs_gotenberg:
                 # Enable PDF for Universal Access for optimal accessibility - default false.
                 pdf_universal_access: null
 
+                # Either intervals or pages. - default None. https://gotenberg.dev/docs/routes#split-chromium
+                split_mode: null
```
Neirda24, line 235: shouldn't this be in a dedicateds key ?

```yaml
split:
    mode: null
```

same as `html`, `url`, etc etc ?

src/DependencyInjection/Configuration.php
```diff
@@ -597,6 +600,7 @@ private function addPdfOfficeNode(): NodeDefinition
         ;
 
         $this->addPdfFormat($treeBuilder->getRootNode());
+        $this->addSplit($treeBuilder->getRootNode());
```
Neirda24, line 603: okay. I understand now : it is both in pdf generators AND in a deidcated route ?

docs/Pdf/builders_api/HtmlPdfBuilder.md
StevenRenaux: will be fixed there
https://github.com/sensiolabs/GotenbergBundle/pull/133

src/DependencyInjection/Configuration.php
StevenRenaux: yep

Neirda24:
Waiting for https://github.com/sensiolabs/GotenbergBundle/pull/133 to be merged

docs/configuration.md
Neirda24: yes indeed. As i saw later : you have `split_*` in `html` and so on : so keep the prefix.
But you also need

```yaml
split:
    split_mode: xxxx
```

on the same level as html. This does not appear in your documentation I think.

Neirda24 (approved):

PR #137, opened 2025-01-08
Add gotenberg output filename

### LINKED ISSUE #91: Send Gotenberg-Output-Filename if fileName is set
Add this header if fileName is set.

See https://gotenberg.dev/docs/routes#output-filename

StevenRenaux:
Close #91

### DIFF
--- a/src/Builder/AsyncBuilderTrait.php
+++ b/src/Builder/AsyncBuilderTrait.php
@@ -56,7 +56,6 @@ trait AsyncBuilderTrait
         }
 
         if (null !== $this->fileName) {
-            // Gotenberg will add the extension to the file name (e.g. filename : "file.pdf" => generated file : "file.pdf.pdf").
             $headers['Gotenberg-Output-Filename'] = $this->fileName;
         }
         $this->client->call($this->getEndpoint(), $this->getMultipartFormData(), $headers);
--- a/src/Builder/DefaultBuilderTrait.php
+++ b/src/Builder/DefaultBuilderTrait.php
@@ -4,6 +4,7 @@ namespace Sensiolabs\GotenbergBundle\Builder;
 
 use Psr\Log\LoggerInterface;
 use Sensiolabs\GotenbergBundle\Client\GotenbergClientInterface;
+use Sensiolabs\GotenbergBundle\Exception\InvalidBuilderConfiguration;
 use Sensiolabs\GotenbergBundle\Exception\JsonEncodingException;
 use Sensiolabs\GotenbergBundle\Formatter\AssetBaseDirFormatter;
 use Sensiolabs\GotenbergBundle\Processor\NullProcessor;
@@ -69,6 +70,10 @@ trait DefaultBuilderTrait
      */
     public function fileName(string $fileName, string $headerDisposition = HeaderUtils::DISPOSITION_INLINE): static
     {
+        if (!preg_match('/\.[^.]+$/', $fileName)) {
+            throw new InvalidBuilderConfiguration(\sprintf('File name "%s" needs to get extension as ".pdf", ".png" or any other valid extension.', $fileName));
+        }
+
         $this->fileName = $fileName;
         $this->headerDisposition = $headerDisposition;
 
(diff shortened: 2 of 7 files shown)


src/Builder/AsyncBuilderTrait.php
```diff
@@ -56,8 +57,9 @@ public function generateAsync(): void
         }
 
         if (null !== $this->fileName) {
-            // Gotenberg will add the extension to the file name (e.g. filename : "file.pdf" => generated file : "file.pdf.pdf").
-            $headers['Gotenberg-Output-Filename'] = $this->fileName;
+            /** @var string $basename Gotenberg will add the extension to the file name */
+            $basename = preg_replace('/\.[^.]+$/', '', $this->fileName);
```
Neirda24, line 61:
```suggestion
            $basename = preg_replace('/\.pdf$/', '', $this->fileName);
```

or something similar. But we shouldn't prevent from having a `.something` at the end. So either we detect that it ends with `.pdf`, `.jpg` or whatever and we raise a warning log. Or we simply drop this part and use the raw property value.

src/Builder/DefaultBuilderTrait.php
```diff
@@ -69,6 +70,10 @@ protected function encodeData(string $key, mixed $value): array
      */
     public function fileName(string $fileName, string $headerDisposition = HeaderUtils::DISPOSITION_INLINE): static
     {
+        if (!preg_match('/\.[^.]+$/', $fileName)) {
+            throw new InvalidBuilderConfiguration(\sprintf('File name "%s" needs to get extension as ".pdf", ".png" or any other valid extension.', $fileName));
```
Neirda24, line 74: actually no. Because whatever we do Gotenberg will add the extension.

StevenRenaux: yeah but needed for the processor https://github.com/StevenRenaux/GotenbergBundle/blob/features/Add-Gotenberg-Output-Filename-if-fileName-is-set/src/Builder/DefaultBuilderTrait.php#L233

src/Builder/AsyncBuilderTrait.php
Jean-Beru: IMO, we should only document that the extension will be added by Gotenberg. We should limit user input modification to avoid any misunderstanding.

Neirda24: hmmm IMO limiting user modification is weirder. We shouldn't validate or modify whatever is input by the developper.

src/Builder/GotenbergFileResult.php
```diff
@@ -62,9 +60,6 @@ public function stream(): StreamedResponse
     {
         $headers = $this->getHeaders();
         $headers->set('X-Accel-Buffering', 'no'); // See https://symfony.com/doc/current/components/http_foundation.html#streaming-a-json-response
-        if (null !== $this->fileName) {
-            $headers->set('Content-Disposition', HeaderUtils::makeDisposition($this->disposition, $this->fileName));
```
Jean-Beru, line 66: StreamedResponse won't have any filename? Maybe we can set this header with `$response->getFileName()`.

src/Builder/DefaultBuilderTrait.php
Neirda24: normally we guess the filename from the response. Maybe there are some edge cases @Jean-Beru ?

src/Builder/GotenbergFileResult.php
StevenRenaux: He will have because of the get the `$headers->all()` from the response (Gotenberg) where you will find the content disposition from Gotenberg and not overrided there by us

Neirda24: Yeah but we do need to override the disposition at the moment. Because Gotenberg doesn't allow us to change it yet. We need to be able to extract the Content-Disposition and override only the disposition part.

src/Builder/AsyncBuilderTrait.php
```diff
@@ -4,6 +4,7 @@
 
 use Sensiolabs\GotenbergBundle\Exception\MissingRequiredFieldException;
 use Sensiolabs\GotenbergBundle\Webhook\WebhookConfigurationRegistryInterface;
+use Symfony\Component\HttpFoundation\File\File;
```
Neirda24, line 7: to be removed ?

src/Builder/DefaultBuilderTrait.php
Jean-Beru: We can extract the filename from the response to pass it to the processor constructor. But it means that the request will be sent before calling `GotenbergFileResult::process` or `GotenbergFileResult::stream` to retrieve headers.

An other solution could be that processors doesn't receive the filename in their constructor. They will have to retrieve it by themselves from a chunk.

src/Builder/GotenbergFileResult.php
Jean-Beru: Didn't say anything, `$headers = $this->getHeaders();` already retrieve Gotenberg response headers.

src/Builder/AsyncBuilderTrait.php
Jean-Beru: I want to limit "input modification", not "limit input" :smile:

src/Builder/DefaultBuilderTrait.php
Neirda24: I need to dig a bit to understand more.

Neirda24:
closing as already in 1.x

PR #141, opened 2025-01-11
Doc/processor

Neirda24:
Closes https://github.com/sensiolabs/GotenbergBundle/issues/140

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,6 @@
         "async-aws/s3": "^2.6",
         "friendsofphp/php-cs-fixer": "^3.41",
         "league/flysystem": "^3.29",
-        "league/flysystem-bundle": "^3.3",
         "phpstan/extension-installer": "^1.3",
         "phpstan/phpstan": "^1.10",
         "phpstan/phpstan-symfony": "^1.3",
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,9 @@
         "symfony/mime": "^6.4 || ^7.0"
     },
     "require-dev": {
+        "async-aws/s3": "^2.6",
         "friendsofphp/php-cs-fixer": "^3.41",
+        "league/flysystem-bundle": "^3.3",
         "phpstan/extension-installer": "^1.3",
         "phpstan/phpstan": "^1.10",
         "phpstan/phpstan-symfony": "^1.3",
--- a/composer.json
+++ b/composer.json
@@ -33,8 +33,10 @@
         "symfony/mime": "^6.4 || ^7.0"
     },
     "require-dev": {
+        "ext-mbstring": "*",
         "async-aws/s3": "^2.6",
         "friendsofphp/php-cs-fixer": "^3.41",
+        "league/flysystem": "^3.29",
         "league/flysystem-bundle": "^3.3",
         "phpstan/extension-installer": "^1.3",
         "phpstan/phpstan": "^1.10",
@@ -58,7 +60,9 @@
     },
     "suggest": {
         "symfony/monolog-bundle": "Enables logging througout the generating process.",
-        "symfony/twig-bundle": "Allows you to use Twig to render templates into PDF",
-        "monolog/monolog": "Enables logging througout the generating process."
+        "symfony/twig-bundle": "Allows you to use Twig to render templates into PDF.",
+        "monolog/monolog": "Enables logging througout the generating process.",
+        "async-aws/s3": "Upload any file to aws s3 compatible endpoints supporting multi part upload without memory overhead.",
+        "league/flysystem-bundle": "Upload any file using this filesystem abstraction package."
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
         "phpstan/phpstan": "^1.10",
         "phpstan/phpstan-symfony": "^1.3",
         "phpunit/phpunit": "^10.4",
-        "shipmonk/composer-dependency-analyser": "^1.7",
+        "shipmonk/composer-dependency-analyser": "^1.8",
         "symfony/framework-bundle": "^6.4 || ^7.0",
         "symfony/http-client": "^6.4 || ^7.0",
         "symfony/monolog-bundle": "^3.10",
--- a/docs/processing.md
+++ b/docs/processing.md
@@ -172,7 +172,7 @@ class SomeService
         private readonly GotenbergPdfInterface $gotenbergPdf,
         
         #[Autowire('%kernel.project_dir%/var/pdf')]
-        private readonly string $kernelProjectDir,
+        private readonly string $pdfStorage,
     ) {}
     
     /**
--- a/docs/processing.md
+++ b/docs/processing.md
@@ -104,11 +104,11 @@ This will return a `SplFileInfo` of the generated file stored at `%kernel.projec
 
 </details>
 
-## `Sensiolabs\GotenbergBundle\Processor\NullProcessor`
+### `Sensiolabs\GotenbergBundle\Processor\NullProcessor`
 
 Empty processor. Does nothing. Returns `null`.
 
-## `Sensiolabs\GotenbergBundle\Processor\TempfileProcessor`
+### `Sensiolabs\GotenbergBundle\Processor\TempfileProcessor`
 
 Creates a temporary file and dump all chunks into it. Return a `ressource` of said `tmpfile()`.
 
@@ -144,7 +144,7 @@ class SomeService
 
 </details>
 
-## `Sensiolabs\GotenbergBundle\Processor\ChainProcessor`
+### `Sensiolabs\GotenbergBundle\Processor\ChainProcessor`
 
 Apply multiple processors. Each chunk will be sent to each processor sequentially. Return an array of values returned by chained processors.
 
@@ -199,7 +199,7 @@ class SomeService
 
 </details>
 
-## `Sensiolabs\GotenbergBundle\Bridge\LeagueFlysystem\Processor\FlysystemProcessor`
+### `Sensiolabs\GotenbergBundle\Bridge\LeagueFlysystem\Processor\FlysystemProcessor`
 
 Upload using the `league/flysystem-bundle` package. Returns a `callable`. This callable will return the uploaded content.
 
@@ -241,7 +241,7 @@ class SomeService
 
 </details>
 
-## `Sensiolabs\GotenbergBundle\Bridge\AsyncAws\Processor\AsyncAwsS3MultiPartProcessor`
+### `Sensiolabs\GotenbergBundle\Bridge\AsyncAws\Processor\AsyncAwsS3MultiPartProcessor`
 
 Upload using the `async-aws/s3` package. Uploads using the [multipart upload](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html) feature of S3. Returns a `AsyncAws\S3\Result\CompleteMultipartUploadOutput` object.
 
@@ -278,7 +278,7 @@ class SomeService
 
 </details>
 
-### Custom processor
+## Custom processor
 
 A custom processor must implement `Sensiolabs\GotenbergBundle\Processor\ProcessorInterface` which require that your `__invoke` method is a `\Generator`. To receive a chunk you must assign `yield` to a variable like so : `$chunk = yield`.
 
--- a/docs/processing.md
+++ b/docs/processing.md
@@ -91,7 +91,7 @@ class SomeService
             ->fileName('my_pdf')
             ->processor(new FileProcessor(
                 new Filesystem(),
-                "{$this->kernelProjectDir}/var/pdf",
+                $pdfStorage,
             ))
             ->generate()
             ->process()
--- a/docs/processing.md
+++ b/docs/processing.md
@@ -81,7 +81,7 @@ class SomeService
         private readonly GotenbergPdfInterface $gotenbergPdf,
         
         #[Autowire('%kernel.project_dir%/var/pdf')]
-        private readonly string $kernelProjectDir,
+        private readonly string $pdfStorage,
     ) {}
     
     public function pdf(): \SplFileInfo
--- a/docs/processing.md
+++ b/docs/processing.md
@@ -3,27 +3,34 @@
 Let's say you want to save the PDF or Screenshot as a file, you will need to use a `Sensiolabs\GotenbergBundle\Processor\ProcessorInterface`.
 To avoid loading the whole file content in memory you can stream it to the browser.
 
-You can also hook on the stream and save the file chunk by chunk. To do so we leverage the [`->stream`](https://symfony.com/doc/current/http_client.html#streaming-responses) method from the HttpClientInterface and use a powerful feature from PHP : [`->send`](https://www.php.net/manual/en/generator.send.php).
+You can also hook on the stream and save the file chunk by chunk. To do so we leverage the [`->stream`](https://symfony.com/doc/current/http_client.html#streaming-responses) method from the HttpClientInterface and use a powerful feature from PHP Generators : [`->send`](https://www.php.net/manual/en/generator.send.php).
 
 ## Using FileProcessor
 
-Useful if you want to store the file.
+Useful if you want to store the file in the local filesystem.
 Example when generating a PDF :
+
 ```php
 use Sensiolabs\GotenbergBundle\GotenbergPdfInterface;
 use Sensiolabs\GotenbergBundle\Processor\FileProcessor;
+use Symfony\Component\DependencyInjection\Attribute\Autowire;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\HttpFoundation\Response;
 
 #[Route(path: '/my-pdf', name: 'my_pdf')]
-public function pdf(GotenbergPdfInterface $gotenbergPdf): Response
-{
+public function pdf(
+    GotenbergPdfInterface $gotenbergPdf,
+    Filesystem $filesystem,
+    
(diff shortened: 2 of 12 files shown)


docs/processing.md
```diff
@@ -0,0 +1,80 @@
...
+
+#[Route(path: '/my-pdf', name: 'my_pdf')]
+public function pdf(GotenbergPdfInterface $gotenbergPdf): Response
+{
+    return $gotenbergPdf->html()
+        //
+        ->fileName('my_pdf')
+        ->processor(new FileProcessor(
+            new Filesystem(),
+            $this->getParameter('kernel.project_dir').'/var/pdf',
```
Jean-Beru, line 26: We could encourage the use of `#[Autowire('%kernel.project_dir%/car/pdf')]`.

docs/processing.md
```diff
@@ -0,0 +1,80 @@
...
+        ))
+        ->generate()
+        ->stream()
+    ;
+}
+```
+
+This will save the file under `%kernel.project_dir%/var/pdf/my_pdf.pdf` once the file has been fully streamed to the browser.
+If you are not streaming to a browser you can still process the file this way :
```
Jean-Beru, line 35: If you are not streaming to a browser, you can still process the file using the `process` method instead of `stream` :

docs/processing.md
```diff
@@ -0,0 +1,80 @@
...
+                $this->getParameter('kernel.project_dir').'/var/pdf',
+            ))
+            ->generate()
+            ->process()
+        ;
+    }
+}
+```
+
+This will return a SplFileInfo of the generated file store at `%kernel.project_dir%/var/pdf/my_pdf.pdf`.
```
Jean-Beru, line 62: Stored

docs/processing.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Other processors
+
+* `Sensiolabs\GotenbergBundle\Processor\ChainProcessor` : Apply multiple processor. Each chunk will be sent to each processor sequentially.
```
Jean-Beru, line 66: Multiple processors

docs/processing.md
```diff
@@ -0,0 +1,80 @@
...
+A custom processor must implement `Sensiolabs\GotenbergBundle\Processor\ProcessorInterface` which require that your `__invoke` method is a `\Generator`. To receive a chunk you must assign `yield` to a variable like so : `$chunk = yield`.
+
+The basic needed code is the following :
+
+```php
+do {
+    $chunk = yield;
+    // do something with it
+} while (!$chunk->isLast());
+```
```
Jean-Beru, line 80: We should also explain that the returned value of the processor will be returned by the `process` method

DanielBadura:
What would be the solution when I want to use it via cli & use e.g. flysystem to save the pdf to another source like a s3 bucket? If I understood the docs correctly, this is right now not possible unless I implement a own `Processor`, like e.g. `StringProcessor` which simply returns the file as a string (I know, this could be consuming alot of memory, but this could be used universally). Other solution would be to implement a `Processor` like the `FileProcessor` but coupled to Flysystem.

Neirda24:
> What would be the solution when I want to use it via cli & use e.g. flysystem to save the pdf to another source like a s3 bucket? If I understood the docs correctly, this is right now not possible unless I implement a own `Processor`, like e.g. `StringProcessor` which simply returns the file as a string (I know, this could be consuming alot of memory, but this could be used universally). Other solution would be to implement a `Processor` like the `FileProcessor` but coupled to Flysystem.

A StringProcessor would eventually cancel all the memory efficiency that this bundle tries to provide. But yes a custom Processor to leverage https://github.com/async-aws/aws/blob/master/src/Integration/Aws/SimpleS3/src/SimpleS3Client.php#L69 for example to upload chunk by chunk. We are planning for a flysystem native integration with this bundle.

Jean-Beru:
> What would be the solution when I want to use it via cli & use e.g. flysystem to save the pdf to another source like a s3 bucket? If I understood the docs correctly, this is right now not possible unless I implement a own `Processor`, like e.g. `StringProcessor` which simply returns the file as a string (I know, this could be consuming alot of memory, but this could be used universally). Other solution would be to implement a `Processor` like the `FileProcessor` but coupled to Flysystem.

You're right, you will have to create your own Processor to stream to your bucket. You can use AsyncAws to make it easier.

There is no such class in the GotenbergBundle yet but it's planned as a Bridge.

Neirda24:
> What would be the solution when I want to use it via cli & use e.g. flysystem to save the pdf to another source like a s3 bucket? If I understood the docs correctly, this is right now not possible unless I implement a own `Processor`, like e.g. `StringProcessor` which simply returns the file as a string (I know, this could be consuming alot of memory, but this could be used universally). Other solution would be to implement a `Processor` like the `FileProcessor` but coupled to Flysystem.

Here is a naive approach for async aws : https://gist.github.com/Neirda24/05e6a65a862ab34854be3b5f9e562420 As @Jean-Beru said we will try to integrate natively with Flysystem and so on.

Neirda24:
and here is a working example for flysystem. https://gist.github.com/Neirda24/f950f1aaa5343c67bcbaeb12ed0cb887 I'll have to investigate more to see if its possible without a tmpfile but I don't see how given the abstraction.

composer.json
```diff
@@ -33,12 +33,16 @@
         "symfony/mime": "^6.4 || ^7.0"
     },
     "require-dev": {
+        "ext-mbstring": "*",
+        "async-aws/s3": "^2.6",
```
Neirda24, line 37: might be worth to lower the requirement

composer.json
```diff
@@ -33,12 +33,16 @@
...
         "friendsofphp/php-cs-fixer": "^3.41",
+        "league/flysystem": "^3.29",
+        "league/flysystem-bundle": "^3.3",
```
Neirda24, line 40: might be worth to lower the requirement

composer.json
```diff
@@ -33,12 +33,16 @@
...
         "phpstan/extension-installer": "^1.3",
         "phpstan/phpstan": "^1.10",
         "phpstan/phpstan-symfony": "^1.3",
         "phpunit/phpunit": "^10.4",
-        "shipmonk/composer-dependency-analyser": "^1.7",
+        "shipmonk/composer-dependency-analyser": "^1.8",
```
Neirda24, line 45: needed to ignore `ext-mbstring` issues

smnandre:
> and here is a working example for flysystem. https://gist.github.com/Neirda24/f950f1aaa5343c67bcbaeb12ed0cb887 I'll have to investigate more to see if its possible without a tmpfile but I don't see how given the abstraction.

What about using kbond filesystem that does all the abstractions for you?

Neirda24:
> > and here is a working example for flysystem. https://gist.github.com/Neirda24/f950f1aaa5343c67bcbaeb12ed0cb887 I'll have to investigate more to see if its possible without a tmpfile but I don't see how given the abstraction.
> 
> What about using kbond filesystem that does all the abstractions for you?

Not sure hw that would help in this case. Flysystem does not expose a way to write by chunk. the version from kbonc suffers the same issues.

src/Processor/AsyncAwsProcessor.php
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    public function __invoke(string|null $fileName): \Generator
+    {
+        if (null === $fileName) {
+            $fileName = uniqid('gotenberg_', true);
+            $this->logger?->debug('{processor}: no filename given. Content will be dumped to "{file}".', ['processor' => self::class, 'file' => $fileName]);
+        }
+
+        $this->logger?->debug('{processor}: starting multi part upload of "{file}".', ['processor' => self::class, 'file' => $fileName]);
```
smnandre, line 33: private readonly LoggerInterface $logger = new NullLogger() 

?

src/Processor/FlysystemProcessor.php
```diff
@@ -0,0 +1,49 @@
...
+        $tmpfile = $tmpfileProcessor->getReturn();
+
+        try {
+            $this->filesystemOperator->writeStream($fileName, $tmpfile);
+
+            $this->logger?->debug('{processor}: content dumped to "{file}".', ['processor' => self::class, 'file' => $fileName]);
+
+            return function () use ($fileName) {
+                return $this->filesystemOperator->read($fileName); // use readStream instead ?
+            };
```
smnandre, line 42: Does not feel optimised

Neirda24: it does nothing unless you can that callable. I was thinking of returning an anonymous class that implements Stringable and returns this call instead to make it easier to use. What do you suggest ?

src/Processor/AsyncAwsProcessor.php
Neirda24: Why ? This works well already. What kind of gain would we have ?

src/Processor/AsyncAwsProcessor.php
```diff
@@ -0,0 +1,118 @@
...
+        private S3Client $s3Client,
+        private string $bucketName,
+        private readonly LoggerInterface|null $logger = null,
+    ) {
+    }
+
+    public function __invoke(string|null $fileName): \Generator
+    {
+        if (null === $fileName) {
+            $fileName = uniqid('gotenberg_', true);
```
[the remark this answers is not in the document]
Neirda24: This will be droped soon I think. I don't see a reason filename should be null. I went along with it in this PR but it will change. Otherwise yes filesystem could be a lead.

src/Processor/AsyncAwsProcessor.php
```diff
@@ -0,0 +1,118 @@
...
+use AsyncAws\S3\S3Client;
+use Psr\Log\LoggerInterface;
+use Sensiolabs\GotenbergBundle\Exception\ProcessorException;
+
+/**
+ * TODO : Might be worth adding "MultiPart" to the name as not all services supports the multi part upload.
+ *
+ * @implements ProcessorInterface<CompleteMultipartUploadOutput>
+ */
+final class AsyncAwsProcessor implements ProcessorInterface
```
Jean-Beru, line 15: IMHO, specific processors like AWS or Flysystem must be in a separated bridge to avoid adding dev-dependencies. It could block the update of GotenbergBundle to Symfony x.y if, for example, FlysystemBundle does not support it yet.

docs/processing.md
```diff
@@ -0,0 +1,90 @@
...
```
StevenRenaux, line 90:
```suggestion
```php
public function __invoke(string|null $fileName): \Generator
{

    do {
        $chunk = yield;
        // do something with it
    } while (!$chunk->isLast());

    // rest of your code
}
```
```

File: docs/processing.md [hunk 14]
```diff
@@ -0,0 +1,90 @@
...
+        ;
+    }
+}
+```
+
+This will return a `SplFileInfo` of the generated file stored at `%kernel.project_dir%/var/pdf/my_pdf.pdf`.
+
+## Other processors
+
+* `Sensiolabs\GotenbergBundle\Processor\AsyncAwsProcessor` : Upload using the `async-aws/s3` package. Uploads using the (multipart upload)[https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html] feature of S3. Returns a `AsyncAws\S3\Result\CompleteMultipartUploadOutput` object.
```
StevenRenaux, line 73: Maybe add an example of it

src/Processor/AsyncAwsProcessor.php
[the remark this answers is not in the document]
StevenRenaux: yeah because of the `getFileName()` who's returned `$this->response->getFileName()`. But I think we can't receive a null or it willcrash with an exception before this part.

src/Processor/AsyncAwsProcessor.php
Neirda24: but then they would have to require it just for one class potentially ? Seems a bit too much don't you think ?

Neirda24: IMO such classes must be fixed on package version instead of framework version. So maybe rename it to include more clearly package name and version of said package ?

Jean-Beru: OK for a `Bridge` folder (like the `Notifier` for example) for now.

To avoid developer to install both Flysystem and AsyncAWS on their local project, we can install these dev dependencies only in the CI (if tests need them).

Jean-Beru: Why not. We should also not rely on the bundle but just the FlySystem lib

docs/processing.md
```diff
@@ -0,0 +1,155 @@
...
+
+TODO
+
+## `Sensiolabs\GotenbergBundle\Bridge\LeagueFlysystem\Processor\FlysystemProcessor`
+
+Upload using the `league/flysystem-bundle` package. Returns a `callable`. This callable will return the uploaded content.
+
+## `Sensiolabs\GotenbergBundle\Bridge\AsyncAws\Processor\AsyncAwsS3MultiPartProcessor`
+
+Upload using the `async-aws/s3` package. Uploads using the (multipart upload)[https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html] feature of S3. Returns a `AsyncAws\S3\Result\CompleteMultipartUploadOutput` object.
```
Jean-Beru, line 130: Wrong format

docs/processing.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+This will save the file under `%kernel.project_dir%/var/pdf/my_pdf.pdf` once the file has been fully streamed to the browser.
+
+</details>
+
+
+
+<details>
+<summary>If you are not streaming to a browser, you can still process the file using the `process` method instead of `stream`</summary>
```
Jean-Beru, line 73: Back quotes seem to be broken in the preview.

docs/processing.md
```diff
@@ -0,0 +1,155 @@
...
+
+## `Sensiolabs\GotenbergBundle\Bridge\LeagueFlysystem\Processor\FlysystemProcessor`
+
+Upload using the `league/flysystem-bundle` package. Returns a `callable`. This callable will return the uploaded content.
+
+## `Sensiolabs\GotenbergBundle\Bridge\AsyncAws\Processor\AsyncAwsS3MultiPartProcessor`
+
+Upload using the `async-aws/s3` package. Uploads using the (multipart upload)[https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html] feature of S3. Returns a `AsyncAws\S3\Result\CompleteMultipartUploadOutput` object.
+
+### Custom processor
```
smnandre, line 132:
```suggestion
## Custom processor
```

docs/processing.md
```diff
@@ -0,0 +1,155 @@
...
+        ;
+    }
+}
+```
+
+This will return a `SplFileInfo` of the generated file stored at `%kernel.project_dir%/var/pdf/my_pdf.pdf`.
+
+</details>
+
+## `Sensiolabs\GotenbergBundle\Processor\NullProcessor`
```
smnandre, line 109:
```suggestion
### `Sensiolabs\GotenbergBundle\Processor\NullProcessor`
```

docs/processing.md
```diff
@@ -0,0 +1,155 @@
...
+
+Empty processor. Does nothing. Returns `null`.
+
+## `Sensiolabs\GotenbergBundle\Processor\TempfileProcessor`
```
smnandre, line 113:
```suggestion
### `Sensiolabs\GotenbergBundle\Processor\TempfileProcessor`
```

smnandre: (etc.)

docs/processing.md
```diff
@@ -0,0 +1,304 @@
...
+    public function __invoke(string|null $fileName): \Generator { /* ... */ } // Implement your own logic
+}
+
+class SomeService
+{
+    public function __construct(
+        private readonly GotenbergPdfInterface $gotenbergPdf,
+        
+        #[Autowire('%kernel.project_dir%/var/pdf')]
+        private readonly string $kernelProjectDir,
```
StevenRenaux, line 175:
```suggestion
        private readonly string $pdfStorage,
```

docs/processing.md
```diff
@@ -0,0 +1,304 @@
...
+     */
+    public function pdf(): array
+    {
+        return $this->gotenbergPdf->html()
+            //
+            ->fileName('my_pdf')
+            ->processor(new ChainProcessor([
+                new FileProcessor(
+                    new Filesystem(),
+                    "{$this->kernelProjectDir}/var/pdf",
```
StevenRenaux, line 189:
```suggestion
                    $pdfStorage,
```

docs/processing.md
```diff
@@ -0,0 +1,304 @@
...
+use Sensiolabs\GotenbergBundle\Processor\FileProcessor;
+use Symfony\Component\Filesystem\Filesystem;
+
+class SomeService
+{
+    public function __construct(
+        private readonly GotenbergPdfInterface $gotenbergPdf,
+        
+        #[Autowire('%kernel.project_dir%/var/pdf')]
+        private readonly string $kernelProjectDir,
```
StevenRenaux, line 84:
```suggestion
        private readonly string $pdfStorage,
```

docs/processing.md
```diff
@@ -0,0 +1,304 @@
...
+    ) {}
+    
+    public function pdf(): \SplFileInfo
+    {
+        return $this->gotenbergPdf->html()
+            //
+            ->fileName('my_pdf')
+            ->processor(new FileProcessor(
+                new Filesystem(),
+                "{$this->kernelProjectDir}/var/pdf",
```
StevenRenaux, line 94:
```suggestion
                $pdfStorage,
```

StevenRenaux (approved):

PR #143, opened 2025-01-14
fix: use valid type for webhook configuration example

### LINKED ISSUE #142: ConfigBuilder warning - Array to string conversion
### Reproducer

bin/console cache:clear

### Description

On clearing cache I get a warning:
`Failed to generate ConfigBuilder for extension Sensiolabs\GotenbergBundle\DependencyInjection\SensiolabsGotenbergExtension: Warning: Array to string conversion`
I only noticed when building in production where warnings are elevated.

The warning seems to come from the example route provided in `Configuration::addWebhookConfigurationNode()`:
`->example([
                        'https://webhook.site/#!/view/{some-token}',
                        ['my_route', ['param1' => 'value1', 'param2' => 'value2']],
                    ])`

This triggers a warning in Symfony's `ConfigBuilderGenerator::getComment()`:

`if (!$node instanceof ArrayNode) {
            foreach ((array) ($node->getExample() ?? []) as $example) {
                $comment .= ' * @example '.$example."\n";
            }`

Is the entire example node meant to be wrapped in quotation marks?

### Gotenberg Version

8.15.3

### Gotenberg Bundle Version

0.3.0

Jean-Beru:
| Gotenberg API version ? |       |
| Issues                  | Fix #142  |

## Description

`example` should be set for `url` and `route` independently. 

Note: `route` example won't be dumped until we specify its children.

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -756,7 +756,7 @@ class Configuration implements ConfigurationInterface
                 ->end()
                 ->arrayNode('route')
                     ->info('Route configuration.')
-                    ->example(['my_route', ['param1' => 'value1', 'param2' => 'value2']])
+                    ->example([['my_route', ['param1' => 'value1', 'param2' => 'value2']]])
                     ->beforeNormalization()
                         ->ifArray()
                             ->then(function (array $v): array {
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -752,9 +752,11 @@ class Configuration implements ConfigurationInterface
             ->children()
                 ->scalarNode('url')
                     ->info('The URL to call.')
+                    ->example('https://webhook.site/#!/view/{some-token}')
                 ->end()
-                ->variableNode('route')
+                ->arrayNode('route')
                     ->info('Route configuration.')
+                    ->example(['my_route', ['param1' => 'value1', 'param2' => 'value2']])
                     ->beforeNormalization()
                         ->ifArray()
                             ->then(function (array $v): array {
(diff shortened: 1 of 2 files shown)


src/DependencyInjection/Configuration.php
```diff
@@ -752,9 +752,11 @@ private function addWebhookConfigurationNode(string $name): NodeDefinition
             ->children()
                 ->scalarNode('url')
                     ->info('The URL to call.')
+                    ->example('https://webhook.site/#!/view/{some-token}')
                 ->end()
-                ->variableNode('route')
+                ->arrayNode('route')
                     ->info('Route configuration.')
+                    ->example(['my_route', ['param1' => 'value1', 'param2' => 'value2']])
```
Neirda24, line 759:
```suggestion
                    ->example([['my_route', ['param1' => 'value1', 'param2' => 'value2']]])
```
I think.

src/DependencyInjection/Configuration.php
```diff
@@ -752,9 +752,11 @@ private function addWebhookConfigurationNode(string $name): NodeDefinition
             ->children()
                 ->scalarNode('url')
                     ->info('The URL to call.')
+                    ->example('https://webhook.site/#!/view/{some-token}')
                 ->end()
-                ->variableNode('route')
+                ->arrayNode('route')
```
Neirda24, line 757: Why array ? It could be a string as well. Or is there already a normalization process to convert from string to array ?

src/DependencyInjection/Configuration.php
Jean-Beru: Good catch

Neirda24 (approved):

Neirda24:
thanks @Jean-Beru

PR #151, opened 2025-02-28
Add `{{ gotenberg_font() }}` twig function

### LINKED ISSUE #148: Automatically add fonts
When we use twig/markdown or whatever. We can usually define our fonts.
But we could add a twig function to this, but we have two possibilities

# First option

This could be with a twig function `gotenberg_font` where we add the fonts to load and automatically generate a `@font-face`

```twig
<style>

    {{ gotenberg_font(['font.ttf','my_font']) }}

    {# And generate that at the end
        @font-face {
            font-family: "my_font";
            src: url("font.ttf");
        }
    #}

    h1 {
        color: red;
        font-family: "my_font";
    }
</style>
```

# Second option

Or we just call the twig function `gotenberg_font`, and this function automatically resolves fonts added from the `asset` or `addAssets` functions.

```php
return $gotenberg
    ->html()
    ->content('content.html.twig')
    ->addAsset('font.ttf')
    ->generate()
;
```

```twig
<style>
    {{ gotenberg_font() }}

    {# And generate that at the end
        @font-face {
            font-family: "my_font";
            src: url("font.ttf");
        }
    #}

    h1 {
        color: red;
        font-family: "my_font";
    }
</style>
```

StevenRenaux:
| Issues                  | Fix #148  |

## Twig file

The `{{ gotenberg_font() }}` function helps generate an `@font-face` 
declaration with the correct asset path expected by gotenberg.

You can provide an absolute path.

### Example

```html
<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="utf-8" />
        <title>PDF with Custom Font</title>
        <style>
            {{ gotenberg_font('fonts/custom-font.ttf', 'MyCustomFont') }}
            h1 {
                color: red;
                font-family: "my_font";
            }
        </style>
    </head>
    <body>
        <h1>This text uses the custom font.</h1>
    </body>
</html>
```

### Output

```html
```html
<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="utf-8" />
        <title>PDF with Custom Font</title>
        <style>
            @font-face {
                font-family: "MyCustomFont";
                src: url("custom-font.ttf");
            }
            h1 {
                color: red;
                font-family: "my_font";
            }
        </style>
    </head>
    <body>
        <h1>This text uses the custom font.</h1>
    </body>
</html>
```

And in your controller nothing needs to be changed.

```php
namespace App\Controller;

use Sensiolabs\GotenbergBundle\GotenbergPdfInterface;

class YourController
{
    public function yourControllerMethod(GotenbergPdfInterface $gotenberg): Response
    {
        return $gotenberg->html()
            ->content('twig_simple_pdf.html.twig')
            ->generate()
            ->stream()
         ;
    }
}
```

## HTML file

If your file is an HTML file (not a Twig template), you can still include 
fonts manually.

The only requirement is that their paths in the HTML file are on the root level.

```html
<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="utf-8" />
        <title>PDF with Custom Font</title>
        <style>
            @font-face {
                font-family: "MyCustomFont";
                src: url("custom-font.ttf");
            }
        </style>
    </head>
    <body>
        <p style="font-family: 'MyCustomFont';">This text uses the custom font.</p>
    </body>
</html>
```

All you need to do is to add the path of the asset file to either 
`assets(...string)` or `addAsset(string)` function.

```php
namespace App\Controller;

use Sensiolabs\GotenbergBundle\GotenbergPdfInterface;

class YourController
{
    public function yourControllerMethod(GotenbergPdfInterface $gotenberg): Response
    {
        return $gotenberg->html()
            ->contentFile('content.html')
            ->assets('fonts/my-font.ttf') // By default, the assets are fetch in the `assets` folder of your application.
            ->generate()
            ->stream()
        ;
    }
}

### CODE REVIEW

File: docs/fonts.md [hunk 1]
```diff
@@ -0,0 +1,140 @@
...
+    <body>
+        <h1>This text uses the custom font.</h1>
+    </body>
+</html>
+```
+
+### Output
+
+```html
+```html
```
Neirda24, line 51: remove one ?

src/Twig/GotenbergRuntime.php
```diff
@@ -25,13 +25,32 @@ public function setBuilder(AbstractChromiumPdfBuilder|AbstractChromiumScreenshot
...
+        $this->addAsset($path, 'gotenberg_asset');
+
+        return basename($path);
+    }
+
+    public function getFont(string $path, string $name): string
+    {
+        $this->addAsset($path, 'gotenberg_font');
+
+        $name = htmlspecialchars($name);
```
Neirda24, line 38: is this needed ? might be optional and fallback on basename ?
also if htmlspecialchars is needed to be sure it is compatible then I'd check if there are any diff between `$name` and `htmlspecialchars($name)` to throw error. Because then I assume that this name is referenced by the developper somewhere else right ?

StevenRenaux: `basename` give us the name with the extension who is not really user friendly I think 🤷‍♂️​​
And the value for `font-family` can sometimes be an other name.

About `htmlspecialchars`, it's more because I put the option `is_safe` into the TwigFunction argument, to avoid the default behavior of escaping. https://twig.symfony.com/doc/3.x/advanced.html#automatic-escaping

But I need, I think, to convert special char if the developper pass some wrong values...

Neirda24: maybe ensure the name is not empty ?

Neirda24 (approved):
Might be worth creating an issue to allow adding fonts from the configuration itself ?

src/Twig/GotenbergRuntime.php
StevenRenaux: You mean empty string? Same for the path argument then?

StevenRenaux:
> Might be worth creating an issue to allow adding fonts from the configuration itself ?

I'm not really in favor of this, because we're back to the other solution. Loop over all the other assets to finally update the template on the fly.

docs/fonts.md
```diff
@@ -0,0 +1,139 @@
+# Working with fonts
+
+You can add fonts in the same way as assets. The `gotenberg_font()` function follows 
+the same path resolution logic as [gotenberg_asset()](assets.md), but it generates an `@font-face`
```
Jean-Beru, line 4:
```suggestion
the same path resolution logic as [gotenberg_asset()](assets.md), but it generates a `@font-face`
```

docs/fonts.md
```diff
@@ -0,0 +1,139 @@
...
+>
+> For header and footer, only fonts installed in the Docker image are loaded - 
+> see the [fonts configuration section](https://gotenberg.dev/docs/configuration#fonts).
+
+## Twig file
+
+The `{{ gotenberg_font() }}` function helps generate an `@font-face` 
+declaration with the correct asset path expected by gotenberg.
+
+You can provide an absolute path.
```
Jean-Beru, line 24: Already said previously (in `# Working with fonts`)

docs/fonts.md
```diff
@@ -0,0 +1,139 @@
+# Working with fonts
+
+You can add fonts in the same way as assets. The `gotenberg_font()` function follows
```
Jean-Beru, line 3: Maybe we can show arguments. Ex: `gotenberg_font(path_to_font, font_family)`. WDYT?

docs/fonts.md
```diff
@@ -0,0 +1,139 @@
...
+    }
+}
+```
+
+## HTML file
+
+If your file is an HTML file (not a Twig template), you can still include 
+fonts manually.
+
+The only requirement is that their paths in the HTML file are on the root level.
```
Jean-Beru, line 98:
```suggestion
The only requirement is that their paths in the HTML file must be on the root level.
```

Jean-Beru (approved):

PR #154, opened 2025-04-11
AssetBaseDirFormatter Trim windows paths

smnandre:

## Description

To be fair, this does not change a thing at runtime, as `$this->baseDir` is only consumed by `Path::join(...)` method (that does this already)... for now.

But for defensive programmation it's also done in the constructor.

Then, to enforce the same behaviour on windows... let's trim both path spearators


Neirda24 (approved):
thanks @smnandre

PR #160, opened 2025-05-07
chore: update readme

94noni:
| Issues                  | Fix readme reading  |

## Description

- move above that the bundle rely on the tool
- swap the caution in favor of a note

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,10 +18,6 @@ Markdown or any Office file. Different options are available depending on the so
 It also helps you to generate, stream and save images locally from URL, HTML and
 Markdown by taking a screenshot.
 
-> [!NOTE]
-> To use this bundle, you first need to install and configure [Gotenberg 8.x](https://gotenberg.dev/docs/getting-started/installation)
-> which is used under the hood.
-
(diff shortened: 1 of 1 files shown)


Jean-Beru:
Hello @94noni,

Thanks for the patch!

I'm wondering if we can move this part before the "with Flex" part since the Flex recipe will add Gotenberg in the Docker compose stack :thinking: 

We may add notes in "with Flex" and "without Flex" parts to explain that the first updates your Docker stack and the second needs a working Gotenberg instance. WDYT?

Neirda24 (approved):
totally makes sense. thank you @94noni

Jean-Beru:
> for a newcomer, seeing the tool on which this bundle rely on top is more relevant 👍🏻

You're right. Especially since this tool deserves to be displayed first! I pushed a new commit. Feel free to review it.

> using it more and more 🚀

:heart:

94noni (commented):
✅ approved!

Jean-Beru:
Thanks @94noni !

Issue #66, opened 2024-06-10
Create a test kit


Neirda24:
The goal is to be able to easily mock / test any application using this bundle without having to mock everything ourselves.

smnandre:
Do you have any spec/draft/shopping list about this?

We may need something similar soonish at work 👼

Neirda24:
* A way to easily test custom builder
* a simple MockGotenbergClient similar to HttpClient one where I can provide a path to a hard coded file.
* Maybe some custom Assert `$this->assertSensioLabsGotenbergGeneratedPdfCount(xxx)`, `$this->assertSensioLabsGotenbergGeneratedFileMatches('/path/to/test-file')`

WDYT ?

smnandre:
I think both are great idea... priority would depend on your short-term goals: 
* more "users" --> let's start with MockClient, and Assert then if possible
* more implementations / builder / processors -> BuilderTestCase 

Clearly, having a MockGotenberg would add great value to the project DX.

Jean-Beru:
> * a simple MockGotenbergClient similar to HttpClient one where I can provide a path to a hard coded file.

Maybe something similar to https://github.com/sensiolabs/GotenbergBundle/pull/118/files#diff-39ed97fae618725a167e6c6cf5924d99853127dcc4fde373a28e9e4493a96a73R1 ? It could allow to mock and make some assertions.

> Maybe some custom Assert $this->assertSensioLabsGotenbergGeneratedPdfCount(xxx), $this->assertSensioLabsGotenbergGeneratedFileMatches('/path/to/test-file')

Like `Mailer` and `Messenger`. We could use the existing DataCollector or throw some Events (DataCollector could also use them)

Jean-Beru:
Fixed by #166 and #171

PR #166, opened 2025-06-11
Test kit

Neirda24:

## Description

* Adds test kit
* Move PHPStan rule
* Fix issues with GotenbergFileResult


src/Test/Builder/GotenbergMockFileResult.php
```diff
@@ -0,0 +1,61 @@
...
+
+        $response = $client->request('GET', '/');
+
+        return new self($client->stream($response), new NullProcessor(), $disposition);
+    }
+
+    /**
+     * @param array<string, array<string>> $headers
+     */
+    public static function fromContent(
```
Jean-Beru, line 49: This code could be in the constructor to avoid "weird" instanciation. Also, `fromFile` could be updated to
```php
return self(file_get_content($path), $filename ?? basename($path), $headers, $disposition);
```

src/Test/Builder/GotenbergBuilderTestCase.php
```diff
@@ -16,7 +16,7 @@
  */
 abstract class GotenbergBuilderTestCase extends TestCase
```
Jean-Beru, line 17: Note for later: this helper should be documented.

src/Test/Builder/GotenbergMockFileResult.php
Neirda24: I think we should not override the constructor if we don't need to. I agree that at the moment we could prevent it from being used by setting the constructor private. As such if we want to open the visbility later on there won't be any BC.

src/Test/Builder/GotenbergBuilderTestCase.php
Neirda24: Yes 👍

src/Test/Builder/GotenbergMockFileResult.php
Jean-Beru: OK let's keep as it. But I still think that we can reduce the `fromFile` to a single line.

src/Builder/Result/GotenbergFileResult.php
```diff
@@ -17,48 +17,38 @@ class GotenbergFileResult extends AbstractGotenbergResult
...
 
     /**
      * @template TNewProcessorResult of mixed = mixed
      *
      * @param ProcessorInterface<TNewProcessorResult> $processor
      *
-     * @return self<TNewProcessorResult>
+     * @phpstan-assert ProcessorInterface<TNewProcessorResult> $this->processor
+     *
+     * @phpstan-this-out self<TNewProcessorResult>
```
StevenRenaux, line 41: What the purpose of this annotation? :eyes:

src/Test/Builder/GotenbergBuilderTestCase.php
Jean-Beru: Issue #167 created.

Jean-Beru:
Does this PR fix #66 ?

StevenRenaux (approved):

src/Test/Builder/GotenbergMockFileResult.php
StevenRenaux: I don't get it @Jean-Beru. What kinf of weird instanciation?

src/Builder/Result/GotenbergFileResult.php
```diff
@@ -17,48 +17,38 @@ class GotenbergFileResult extends AbstractGotenbergResult
...
     /**
-     * @param ProcessorInterface<TProcessorResult> $processor
+     * @phpstan-assert ProcessorInterface<TProcessorResult> $this->processor
      */
     public function __construct(
-        ResponseInterface $response,
         private readonly ResponseStreamInterface $stream,
-        private readonly ProcessorInterface $processor,
+        private ProcessorInterface $processor,
         private readonly string $disposition,
```
Jean-Beru: If we follow this logic, disposition could also be overridden, right?

src/Test/Builder/GotenbergMockFileResult.php
Jean-Beru: I don't see any good use case for `new GotenbergMockFileResult($trickyStream, $processorButWhy, $disposition)`.

Neirda24: made the constructor private. I think it prevents people from using the new syntax. Makes intent clearer and allwos us to open the constructor later on if we do find a concrete usecase

src/Builder/Result/GotenbergFileResult.php
Neirda24: done

Jean-Beru (approved):

PR #168, opened 2025-06-12
Add dagger

Neirda24:
| New feature ?           | yes  |
| BC break ?              | yes  |

## Description

Make tests reproductible locally using dagger.
Open the possibility to create end to end tests with different Gotenberg version

How to try this out ?
1. Install latest dagger from https://docs.dagger.io/install
2. checkout this branch and run `dagger develop` at the root of the bundle.
3. then you can list all available functions with `dagger functions` (for example you could run `dagger call tests-matrix`)

See the `CONTRIBUTING.md` on this branch to see more

For better performance you can try an experimental cache with : 
`export _EXPERIMENTAL_DAGGER_CACHE_CONFIG="mode=max,image-manifest=true,oci-mediatypes=true,type=registry,ref=docker.company.com/cbochs/webapp-cache"`

before running the `dagger call`

---

Here is the list of all `dagger call` functions you can do :

```shell
$ dagger functions
Name            Description
generate-docs   Generates documentation and returns the Directory to export locally.
test            Provide a container with all dependencies installed and ready to run tests.
tests-matrix    Execute all tests within matrix (PHP version, Symfony version).
```

and here is the list of all tests available in `dagger call test` :

```shell
$ dagger functions test # e.g.: dagger call test phpunit stdout
Name                    Description
all                     Run all tests.
php-cs-fixer            Validate PHP-CS-Fixer and returns the container it ran in.
phpstan                 Run PHPStan and returns the container it ran in.
phpunit                 Run phpunit tests and returns the container it ran in.
terminal                Get the container for tests.
validate-dependencies   Validate composer dependencies and returns the container it ran in.
```


.dagger/src/GotenbergBundle.php
```diff
@@ -0,0 +1,186 @@
...
+        string $phpVersion = '8.4',
+    ): Container {
+        $aptCache = dag()->cacheVolume("apt-cache-{$phpVersion}");
+        $composerBin = dag()->container()->from('composer/composer')->file('/usr/bin/composer');
+
+        return dag()
+            ->container()
+            ->from("php:{$phpVersion}")
+            ->withWorkdir('/GotenbergBundle')
+            ->withMountedDirectory('/GotenbergBundle', $source)
```
Jean-Beru: Are steps ordered? If so, it may be interesting to `apt install` before mounting the source folder.

Neirda24: yes indeed.

.dagger/README.md
```diff
@@ -0,0 +1,447 @@
+# Dagger PHP SDK
```
Jean-Beru: Is this readme useful since we already initialized Dagger ?

.dagger/composer.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "daggermodule/gotenbergbundle",
+  "description": "",
+  "version": "1.0.0",
+  "minimum-stability": "dev",
+  "license": "proprietary",
```
Jean-Beru: MIT?

.dagger/src/GotenbergBundle.php
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    #[DaggerFunction]
+    #[Doc('Runs PHP CS Fixer and returns the container in which it ran.')]
+    public function testCsFixer(
+        #[DefaultPath('.')]
+        Directory $source,
+
+        string $phpVersion = '8.4',
+        string $symfonyVersion = '7.3',
```
Jean-Beru: Is it possible to add a `bool $fix = false` argument to allow fixing CS using Dagger?

.dagger/src/GotenbergBundle.php
```diff
@@ -0,0 +1,242 @@
...
+    #[DaggerFunction]
+    #[Doc('Execute all tests within matrix (PHP version, Symfony version).')]
+    #[ReturnsListOfType('string')]
+    public function testsMatrix(
+        #[DefaultPath('.')]
+        Directory $source,
+    ): array {
+        $result = [];
+        foreach (['8.2', '8.3', '8.4'] as $phpVersion) {
+            foreach (['6.4.*', '7.2.*', '7.3.*'] as $symfonyVersion) {
```
Jean-Beru: Why 7.2 and not 7.1?

.dagger/src/GotenbergBundle.php
```diff
@@ -0,0 +1,242 @@
...
+                $result[] = $this->testPhpunitUnit($source, $phpVersion, $symfonyVersion, $symfonyContainer)->stdout();
+
+                $result[] = '  >> Validating dependencies...';
+                $result[] = $this->testValidateDependencies($source, $phpVersion, $symfonyVersion, $symfonyContainer)->stdout();
+
+                $result[] = '  >> Checking code style...';
+                $result[] = $this->testCsFixer($source, $phpVersion, $symfonyVersion, $symfonyContainer)->stdout();
+
+                $result[] = '  >> Checking phpstan...';
+                $result[] = $this->testPhpstan($source, $phpVersion, $symfonyVersion, $symfonyContainer)->stdout();
```
Jean-Beru: Can we reuse the `test` fonction here instead of copying steps?

.dagger/src/GotenbergBundle.php
```diff
@@ -0,0 +1,242 @@
...
+            ->withMountedCache('/var/cache/apt/archives', $aptCache)
+            ->withExec(['apt', 'update'])
+            ->withExec(['apt', 'install', '--yes',
+                'git',
+                'zip',
+            ])
+            ->withMountedFile('/usr/bin/composer', $composerBin)
+            ->withEnvVariable('COMPOSER_ALLOW_SUPERUSER', '1')
+            ->withWorkdir('/GotenbergBundle')
+            ->withMountedDirectory('/GotenbergBundle', $source)
```
Jean-Beru, line 60: PHP extensions do not need to be installed? Maybe using Pie?

CONTRIBUTING.md
```diff
@@ -40,10 +40,56 @@ $ composer install
...
+Make sure you have [dagger >= v0.18.10](https://docs.dagger.io/install) installed. Then run
+
+```shell
+$ dagger develop
+```
+
 ### Run Tests
 
 ```shell
-$ ./vendor/bin/phpunit
```
Jean-Beru, line 46: We could keep the "classic" way for people who does not to install Dagger

.dagger/src/GotenbergBundle.php
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    #[DaggerFunction]
+    #[Doc('Returns a PHP container.')]
+    public function phpContainer(
+        #[DefaultPath('.')]
+        Directory $source,
+
+        string $phpVersion = '8.4',
+    ): Container {
```
charjr: Correct me if I'm wrong, but this looks like an internal method. If you do not expect anyone to be typing `dagger call php-container` then you can remove the attributes and make it private.

```suggestion
    private function phpContainer(
        Directory $source, 
        string $phpVersion = '8.4'
    ): Container {
```


Dagger only cares what goes in and what comes out. 
Once you're into a DaggerFunction, you can do whatever you want, normal PHP rules apply. You can use private methods, other classes, whatever; as long as you return what it expects at the end.

.dagger/src/GotenbergBundle.php
```diff
@@ -0,0 +1,242 @@
...
+    #[DaggerFunction]
+    #[Doc('Returns a PHP container with symfony set to the desired version.')]
+    public function symfonyContainer(
+        #[DefaultPath('.')]
+        Directory $source,
+
+        string $phpVersion = '8.4',
+        string $symfonyVersion = '7.3',
+        Container|null $phpContainer = null,
+    ): Container {
```
charjr, line 69:
```suggestion
    private function symfonyContainer(
        Directory $source,
        string $phpVersion = '8.4',
        string $symfonyVersion = '7.3',
        Container|null $phpContainer = null,
    ): Container {
```

Similar to the `phpContainer` function, if this is an internal method, no need to expose it to Dagger.

.dagger/src/GotenbergBundle.php
```diff
@@ -0,0 +1,242 @@
...
+            ->withExec(['composer', 'global', 'require', 'symfony/flex'])
+            ->withExec(['composer', 'update'])
+        ;
+    }
+
+    #[DaggerFunction]
+    #[Doc('Runs PHPUnit unit tests and returns the container in which it ran.')]
+    public function testPhpunitUnit(
+        #[DefaultPath('.')]
+        Directory $source,
```
charjr: You have a lot of `testsThing` methods that all take the same arguments.

You might like the idea of extracting them into a subclass.

In this main class having a `test` method that returns a secondary Dagger Object:

```php
#[DaggerFunction]
public function test(
    #[DefaultPath('.')]
    Directory $source,
    string $phpVersion = '8.4',
    string $symfonyVersion = '7.3',
    Container|null $symfonyContainer = null,
): TestsGotenbergBundle {
    $symfonyContainer ??= $this
        ->symfonyContainer($source, $phpVersion, $symfonyVersion);

    return new TestsGotenbergBundle($symfonyContainer);
}
```

Then

```php
#[DaggerObject]
class TestsGotenbergBundle
{
    public function __construct(
        private Container $symfonyContainer,
    ) {}

    #[DaggerFunction]
    public function validateDependencies(): Container

    #[DaggerFunction]
    public function phpunitUnit(): Container
    
    #[DaggerFunction]
    public function phpCsFixer(): Container

    #[DaggerFunction]
    public function phpstan(): Container

    #[DaggerFunction]
    public function tests(): string
}
```

Then back in your main class again you can refactor your `testsMatrix`:

```php
    #[DaggerFunction]
    #[ReturnsListOfType('string')]
    public function testsMatrix(
        #[DefaultPath('.')]
        Directory $source,
    ): array {
        $result = [];
        foreach (['8.2', '8.3', '8.4'] as $phpVersion) {
            foreach (['6.4.*', '7.2.*', '7.3.*'] as $symfonyVersion) {
                $result[] = $this->test($source, $phpVersion, $symfonyVersion)->tests();
            }
        }
       
        // ...
```

charjr (commented):
Left a few possible DX adjustments you could consider 🙂

.dagger/README.md
Neirda24: auto generated. Didn't tried to remove it yet. Will do and let you know.

.dagger/composer.json
Neirda24: maybe. again : auto generated

.dagger/src/GotenbergBundle.php
Neirda24: well... it isn't that simple 😅... Dagger executes in a sandboxed environement. thus any changes like "fix" would not apply to your local filesystem. You'd then need to export the directory which implies to do more steps... I suggests dagger is only used for tests. Even the generate doc I think I should restrict it to "test" things only.

.dagger/src/GotenbergBundle.php
Neirda24: because of https://symfony.com/releases#symfony-releases-calendar

.dagger/src/GotenbergBundle.php
Neirda24: yes. But I didn't to try to have a cache for symfony container. Will asks if it makes any differences

.dagger/src/GotenbergBundle.php
Neirda24: which extensions ?

CONTRIBUTING.md
Neirda24: yes but then we could end up with "I didn't test that way" and break the "all tests run the same everywhere." WDYT ?

.dagger/src/GotenbergBundle.php
Neirda24: Yes I thought about dagger query and stuff like this where there seems to be a best practice about small atomic functions that can be composed in any way. Maybe this does not apply to "projects" but only reusable functions ?

.dagger/README.md
Neirda24: removed

.dagger/composer.json
Neirda24: updated

.dagger/src/GotenbergBundle.php
Neirda24: done

.dagger/src/GotenbergBundle.php
Neirda24: updated to private

.dagger/src/GotenbergBundle.php
Neirda24: updated to use a TestGotenbergBundle object indeed. DX is way better. thank you

.dagger/src/GotenbergBundle.php
Jean-Beru: json and mbstring. Are they already present in the base image?

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,72 @@
...
+jobs:
+
+  test:
+    name: 'Build and Test - PHP ${{ matrix.php }} - Symfony ${{ matrix.symfony-version }}'
+    runs-on: 'ubuntu-latest'
+    strategy:
+      matrix:
+        symfony-version:
+          - '6.4.*'
+          - '7.2.*'
```
Jean-Beru: Needed since 7.3 release?

CONTRIBUTING.md
Jean-Beru: Maybe we can add a "not recommended" warning and place them in a collapsed div which explain that these scripts exists but Dagger has a lot of advantages.

Some people can't/don't want to install Dagger, and so Docker, on their machine.

CONTRIBUTING.md
```diff
@@ -61,27 +112,27 @@ Maintain high code quality by following these steps before submitting a pull req
 Check your code for style violations:
 
 ```shell
-$ PHP_CS_FIXER_IGNORE_ENV=1 ./vendor/bin/php-cs-fixer check --diff
+$ ./vendor/bin/php-cs-fixer check --diff
```
Jean-Beru, line 119: Why not use Dagger here?

CONTRIBUTING.md
```diff
@@ -61,27 +112,27 @@ Maintain high code quality by following these steps before submitting a pull req
...
 ```
 
 Eventually, you can fix the issues automatically:
 
 ```shell
-$ PHP_CS_FIXER_IGNORE_ENV=1 ./vendor/bin/php-cs-fixer fix --diff
+$ ./vendor/bin/php-cs-fixer fix --diff
```
Jean-Beru, line 125: Dagger can also be used here with an additional argument. WDYT ?

Neirda24: I am not confortable to use dagger for this yet... It won't affect the current files. At best you get an `Directory` object from dagger with modified files. Which you can then export (through the cli only though). Doing so will overrides existing files and creates new ones. but it won't delete files that needs to be deleted for example. Yes with cs-fixer no need. But for doc generation for example we might need this kind of things.

All things put together I feel like dagger is not yet suited for file generation / modification.

CONTRIBUTING.md
Neirda24: yes clearly. I created it. Just needs to update the documentation

CONTRIBUTING.md
Neirda24: ok. Will add a `With dagger` and a `Without dagger`

.github/workflows/unit-tests.yml
Neirda24: updated to have all symfony versions currently maintained / in development and their respective PHP versions officially supported. Let me know. (even added symfony 7.4, 8.0 and PHP 8.5)

.dagger/src/GotenbergBundle.php
```diff
@@ -0,0 +1,151 @@
...
+use function Amp\Future\await;
+use function Dagger\dag;
+
+#[DaggerObject]
+#[Doc('Module for GotenbergBundle')]
+class GotenbergBundle
+{
+    private const SYMFONY_VERSIONS = [
+        '6.4.*' => ['8.1', '8.2', '8.3', '8.4', '8.5-rc'],
+        '7.2.*' => ['8.2', '8.3', '8.4', '8.5-rc'],
```
Jean-Beru: Not needed since august 1st :-P

.github/workflows/static.yml
```diff
@@ -17,32 +17,54 @@ jobs:
...
 
   phpstan:
-    name: 'PHPStan'
+    name: 'PHPStan - PHP ${{ matrix.php }} - Symfony ${{ matrix.symfony-version }}'
     runs-on: 'ubuntu-latest'
+    strategy:
+      matrix:
+        symfony-version:
+          - '6.4.*'
+          - '7.2.*'
```
Jean-Beru: Not needed anymore: https://symfony.com/releases/7.2

.github/workflows/unit-tests.yml
Jean-Beru: IMO, we can add SF 7.4 and PHP 8.5 (is it possible to make them not mandatory?). As previously mentioned in a previous comment, 7.2 is not maintained anymore.

CONTRIBUTING.md
Jean-Beru: Got it, I forgot this point!

.github/workflows/static.yml
Neirda24: yes. I am working on something locally to dynamically fetch the matrix so we don't have to maintain it anymore.

Jean-Beru: A PR every 6 months should be enough :stuck_out_tongue:

.dagger/src/GotenbergBundle.php
```diff
@@ -0,0 +1,144 @@
...
+        ;
+    }
+
+    private function getMatrix(): \Generator
+    {
+        /** @var list<array{name: string, symfony-version: string, php: string, 'allow-failure': bool}> $matrix */
+        $matrix = json_decode(file_get_contents(__DIR__.'/matrix-versions.json'), associative: true);
+
+        foreach ($matrix as $row) {
+            yield $row['name'] => [$row['symfony-version'], $row['php']];
```
Jean-Beru: We can remove the key since it's unused.

.github/workflows/static.yml
```diff
@@ -10,39 +10,65 @@ permissions:
   contents: 'read'
 
 jobs:
+  setup:
+    runs-on: 'ubuntu-latest'
```
Jean-Beru, line 15: With `name: Setup matrix` ?

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+    branches: ['main', '1.x']
+  schedule:
+    - cron: '0 0 * * *'
+
+permissions:
+  contents: 'read'
+
+jobs:
+  setup:
+    runs-on: 'ubuntu-latest'
```
Jean-Beru, line 17: Missing name too

CONTRIBUTING.md
```diff
@@ -61,27 +114,33 @@ Maintain high code quality by following these steps before submitting a pull req
 Check your code for style violations:
 
 ```shell
-$ PHP_CS_FIXER_IGNORE_ENV=1 ./vendor/bin/php-cs-fixer check --diff
+$ ./vendor/bin/php-cs-fixer check --diff
+$ # or using dagger
+$ dagger call test php-cs-fixer
```
Jean-Beru: Can we invert them to have something like following examples:

```suggestion
$ dagger call test php-cs-fixer
$ # or without dagger
$ ./vendor/bin/php-cs-fixer check --diff
```

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,77 @@
...
+            ~/.cache/dagger
+          key: ${{ runner.os }}-dagger-${{ hashFiles('dagger.json', '.dagger/src/**') }}
+          restore-keys: |
+            ${{ runner.os }}-dagger-
+
+      - name: 'Validate dependencies'
+        uses: 'dagger/dagger-for-github@8.0.0'
+        with:
+          version: "latest"
+          args: "test --symfony-version '^7.0' --php-version '8' validate-dependencies stdout"
```
Jean-Beru: Are versions needed ?

Neirda24: required no. But it makes it explicit. How would you change it ?

Jean-Beru: Like in the `CONTRIBUTING.md` file: 

```bash
dagger call test validate-dependencies
```

Jean-Beru (approved):

PR #172, opened 2025-06-23
Update GotenbergRuntime visibility

StevenRenaux:

## Description

When I make a custom builder using twig I need to use `GotenbergRuntime` for my tests.


Neirda24:
I have mixed feelings about this... It means that if we change the runtime API it causes a BC break (to remove it from internal). IMO the runtime provides features taht are in the BC policy but not the runtime itself... What about keeping it as internal but with a phpdoc explaining that it is safe to use in case of external builders ?

StevenRenaux:
@Neirda24 :arrow_up:

Neirda24:
@Jean-Beru WDYT ?

Jean-Beru (approved):
LGTM

Neirda24 (approved):

PR #173, opened 2025-07-09
Add Version endpoint

Neirda24:

## Description

TODO

- [x] Create tests on Version / VersionFetcher / etc etc
- [x] Be able to use multiple versions in tests
- [ ] Add log / exception when an option is not available / badly formatted depending on the version


src/Test/Builder/GotenbergMockFileResult.php
```diff
@@ -30,7 +30,7 @@ public static function fromFile(
         string $disposition = HeaderUtils::DISPOSITION_INLINE,
     ): self {
         $client = new MockHttpClient([
-            MockResponse::fromFile($path, [
+            new MockResponse(file_get_contents($path) ?: throw new \InvalidArgumentException(\sprintf('File not found: "%s".', $path)), [
```
Neirda24, line 33: had to change this @Jean-Beru because not supported in symfony 6.4 (see https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/HttpClient/Response/MockResponse.php it is absent)

src/Gotenberg.php
```diff
@@ -3,11 +3,14 @@
...
 
 use Psr\Container\ContainerInterface;
+use Sensiolabs\GotenbergBundle\Model\Version;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
 
 final class Gotenberg implements GotenbergInterface
 {
     public function __construct(
         private readonly ContainerInterface $container,
+        private readonly HttpClientInterface $client,
```
Jean-Beru: Is it possible to add an optional argument to set version and use the API only if needed? 
```php
    public function __construct(
        private readonly ContainerInterface $container,
        private readonly HttpClientInterface $client,
        private readonly null|string $version = null,
    )
```

Or using a \Closure created from `HttpClientInterface::version` to make it lazy?

```php

    public function __construct(
        private readonly ContainerInterface $container,
        private readonly \Closure|string $version,
    )
```

Neirda24: hmmm Closure is not my style as it blurs the "how" things get done. So maybe ? Otherwise I could create a VersionFetcherInterface with 2 implementation. One dynamic through HttpClient and one Static. That way it can : be lazy AND be removed if the static one is set during compile time. WDYT ?

Neirda24: updated with a new approach to be bale to either have it static or dynamic. What bother me still is :
* what if the versions do not match ? Nothing ? how does doctrine behave in such cases ?
* would be great to put the http version in cache because it won't change that often.

src/DataCollector/GotenbergDataCollector.php
```diff
@@ -23,6 +24,7 @@ final class GotenbergDataCollector extends DataCollector implements LateDataColl
      * @param array<mixed>                                      $defaultOptions
      */
     public function __construct(
+        private readonly GotenbergInterface $gotenberg,
```
Jean-Beru: `VersionFetcherInterface` should be enough, right?

src/Gotenberg.php
Jean-Beru: > how does doctrine behave in such cases ?

The priority goes to the version set in configuration. See https://github.com/doctrine/dbal/blob/acb68b388b2577bb211bb26dc22d20a8ad93d97d/src/Connection.php#L188 

> would be great to put the http version in cache because it won't change that often.

We can inject a `StaticVersionFetcher` instance if the URL contains the version. This value will be put in the Symfony's cache.

Neirda24: yeah but I don't think we should add it to the URL. or we have to make sure it does not conflict with true query strings that might be sent to gotenberg. In our case it is a real URL not a DSN (actually we should probably rename the env variable). I thought about decorating the Client with a Psr cache (optional dependency) to put it to cache. Reminder that this is called only during dbeug at the moment.

src/DataCollector/GotenbergDataCollector.php
Neirda24: indeed. will update tmorrow

src/Gotenberg.php
```diff
@@ -20,4 +23,9 @@ public function screenshot(): GotenbergScreenshotInterface
     {
         return $this->container->get(GotenbergScreenshotInterface::class);
     }
+
+    public function version(): Version
```
Jean-Beru: Is it needed since `VersionFetcherInterface` is used in the DataCollector?

Neirda24: Up to debate... I though it would be worth adding it to the global API. Maybe not ? WDYT ?

Jean-Beru: IMO, not for now without any concrete usecase.

Neirda24: done

Jean-Beru (approved):

PR #182, opened 2025-08-21
feat: Add PhpDoc check

### LINKED ISSUE #159: [CI] Make sure all links in the documentation / phpdoc are still valid
Run a test periodically that find all http links in the source code of this bundle and check if :
* the URL returns a 200 
* if an anchor is provided, check that the loaded page do have an HTML with that ID

The workflow should run :
* on new version of this bundle
  * comment on the pull request
* periodically to make sure we don't have broken links
  * create an issue

StevenRenaux:
| Issues                  | Fix #159  |

image


docs/CheckPhpDoc.php
```diff
@@ -0,0 +1,103 @@
...
+    private function getPhpFiles(array $paths): array
+    {
+        $files = [];
+
+        foreach ($paths as $path) {
+            $iterator = new RecursiveIteratorIterator(new RecursiveDirectoryIterator($path));
+
+            foreach ($iterator as $file) {
+                if ($file->isFile() && pathinfo($file, \PATHINFO_EXTENSION) === 'php') {
+                    $files[] = $file->getPathname();
```
Neirda24: could use a generator here ?

docs/CheckPhpDoc.php
```diff
@@ -0,0 +1,103 @@
...
+                }
+            }
+        }
+
+        return $files;
+    }
+
+    private function extractSeeUrls(string $file): array
```
Neirda24: there may be links in `docs` and in `->info`. WDYT ?

docs/CheckPhpDoc.php
```diff
@@ -0,0 +1,103 @@
...
+        $externalDoc = @file_get_contents($url);
+        if (!$externalDoc) {
+            throw new RuntimeException('Unable to read document');
+        }
+
+        $anchor = strstr($url, '#');
+        if ($anchor) {
+            $id = str_replace('#', '', $anchor);
+
+            if (str_contains($externalDoc, 'id="'.$id.'"') || str_contains($externalDoc, "name='{$id}'")) {
```
Neirda24:
```suggestion
            if (str_contains($externalDoc, "id=\"{$id}\"") || str_contains($externalDoc, "name=\"{$id}\"")) {
```
?

docs/CheckPhpDoc.php
```diff
@@ -0,0 +1,103 @@
...
+    {
+        $content = file_get_contents($file);
+        preg_match_all('/@see\s+(https?:\/\/[^\s\*]+)/', $content, $matches);
+
+        return $matches[1] ?? [];
+    }
+
+    private function checkUrl(string $url): void
+    {
+        $externalDoc = @file_get_contents($url);
```
Neirda24: since we already have a dependency on console. why not add HttpClient & WebCrawler ? (as dev dependencies of course unless already required)

docs/CheckPhpDoc.php
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+}
+
+$application = new Application();
+$application->register('check')
+    ->setCode(function (OutputInterface $output, SymfonyStyle $io) {
+        $checkPhpDoc = new CheckPhpDoc();
+
+        $buildersDir = [
```
Neirda24: so not cheking everyhting ?

docs/CheckPhpDoc.php
```diff
@@ -0,0 +1,103 @@
...
+
+        try {
+            $checkPhpDoc->checkAllSeeUrls($buildersDir, $output);
+        } catch (RuntimeException $e) {
+            $io->error($e->getMessage());
+
+            return Command::FAILURE;
+        }
+
+        $io->success('All external docs are availble.');
```
Neirda24:
```suggestion
        $io->success('All external links are valid.');
```

docs/CheckPhpDoc.php
StevenRenaux: No `->info` anymore, thanks to the traits and our dynamic configuration injection.

docs/ValidateUrlDoc.php
```diff
@@ -0,0 +1,124 @@
...
+
+    public function checkAllUrls(OutputInterface $output): void
+    {
+        $progressBar = new ProgressIndicator($output);
+        $progressBar->start('Processing...');
+
+        $urls = [];
+        $files = $this->getFiles(self::DIR_AND_FILES_TO_CHECK);
+        foreach ($files as $file) {
+            array_push($urls, ...$this->extractUrls($file));
```
Neirda24, line 35: easier way to do this would be to leverage AppendIterator

docs/ValidateUrlDoc.php
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        $urls = array_unique($urls);
```
Neirda24, line 38: will not work if using AppendIterator

docs/CheckPhpDoc.php
```diff
@@ -0,0 +1,124 @@
...
+            'md' => preg_match_all('/\[[^\]]+\]\((https?:\/\/[^\s\*)]+)\)/', $content, $matches),
+        };
+
+        return $matches[1] ?? [];
+    }
+
+    private function checkContentResponse(string $url, string $content): void
+    {
+        $crawler = new Crawler($content);
+        $anchor = strstr($url, '#');
```
Neirda24: use parse_url ?

StevenRenaux: Nice catch, never used it

docs/ValidateUrlDoc.php
StevenRenaux: Maybe let's keep as array, I prefer to not call multiple times same URL

.github/workflows/url-check.yml
```diff
@@ -0,0 +1,25 @@
...
+        uses: 'shivammathur/setup-php@v2'
+        with:
+          php-version: '8.3'
+          coverage: 'none'
+
+      - name: 'Install dependencies'
+        uses: 'ramsey/composer-install@v3'
+
+      - name: 'Check links into PhpDoc'
+        run: 'php ./docs/CheckPhpDoc.php'
```
Jean-Beru:
```suggestion
      - name: 'Check links into PhpDoc'
        uses: 'dagger/dagger-for-github@8.0.0'
        with:
          version: "latest"
          args: "checkUrlDocs"
        env:
          DAGGER_NO_NAG: '1'
```
?

docs/ValidateUrlDoc.php
```diff
@@ -0,0 +1,125 @@
...
+
+        $client = HttpClient::create();
+        $allResponses = [];
+        foreach ($urls as $url) {
+            $allResponses[] = $client->request('GET', $url);
+        }
+
+        foreach ($client->stream($allResponses) as $response => $chunk) {
+            if ($chunk->isLast()) {
+                $this->checkContentResponse($response->getInfo('url'), $response->getContent());
```
Jean-Beru: We could also test status code before.

StevenRenaux: Is it necessary, because `$response->getContent()` will throw an exceptions on 3/4/5xx status codes?

Jean-Beru: Also :smile: 

Checking status code will allow to know if the error comes from a 404 or a missing anchor.

StevenRenaux: Done and about the check for an existing anchor or not  it's done there
https://github.com/StevenRenaux/GotenbergBundle/blob/features/Add-phpdoc-links-check/docs/ValidateUrlDoc.php#L103

dagger.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "GotenbergBundle",
-  "engineVersion": "v0.18.14",
+  "engineVersion": "v0.18.17",
```
Jean-Beru: Is it needed?

docs/ValidateUrlDoc.php
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+
+    private function extractUrls(SplFileInfo $file): array
+    {
+        $content = file_get_contents($file->getPathname());
+
+        match ($file->getExtension()) {
+            'php' => preg_match_all('/(?:@see\s+|->info\([^)]*)(https?:\/\/[^\s\'")]+)(?:[^)]*\))?/', $content, $matches),
+            'md' => preg_match_all('/\[[^\]]+\]\((https?:\/\/[^\s\*)]+)\)/', $content, $matches),
```
Jean-Beru, line 115: Can you add a comment with a link to regex101 to make it clearer?

docs/pdf/office-builder.md
```diff
@@ -64,41 +64,41 @@ class YourController
...
+[addOriginalDocumentAsStream](#addOriginalDocumentAsStream)
 [downloadFrom](#downloadFrom)
 [doNotUpdateIndexes](#doNotUpdateIndexes)
 
 ### Formatting
-[metadata](#metadata)  
-[addMetadata](#addMetadata)  
-[pdfFormat](#pdfFormat)  
-[pdfUniversalAccess](#pdfUniversalAccess)  
-[convertOooTargetToPdfTarget](#convertOooTargetToPdfTarget)
```
Jean-Beru: Should be reverted to preserve line breaks. Maybe adding a `\` at the end of the line would be a better alternative?

.dagger/src/TestsGotenbergBundle.php
```diff
@@ -80,6 +80,16 @@ public function phpCsFixer(): string
         ;
     }
 
+    #[DaggerFunction]
+    #[Doc('Validate all URL docs.')]
+    public function ValidateUrlDoc(): string
```
Neirda24:
```suggestion
    public function validateUrlDoc(): string
```

docs/ValidateUrlDoc.php
```diff
@@ -0,0 +1,134 @@
...
+            $allResponses[] = $client->request('GET', $url);
+        }
+
+        foreach ($client->stream($allResponses) as $response => $chunk) {
+            if ($chunk->isLast()) {
+                $statusCode = $response->getStatusCode();
+                $url = $response->getInfo('url');
+
+                if (200 !== $statusCode) {
+                    throw new RuntimeException("HTTP {$statusCode} error for: {$url}");
```
Neirda24: insteadof exception : error report ?

Neirda24 (approved):

docs/ValidateUrlDoc.php
Jean-Beru: \+ exit code 1 ? The CI should fail if any URL is wrong.

Neirda24: yes. But I think having a unique list of all the URLs taht failed would be better instead of stopping at first issue.

Jean-Beru: Of course :slightly_smiling_face:

.github/workflows/url-test.yml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: 'actions/checkout@v4'
+
+      - name: 'Validate URLS into PhpDoc'
+        uses: 'dagger/dagger-for-github@8.0.0'
+        with:
+          version: "latest"
+          args: "test validate-url-doc"
+        env:
```
Neirda24, line 25: check other files. This part was updated a bit.

.github/workflows/url-test.yml
```diff
@@ -0,0 +1,25 @@
...
+    name: 'Validate URL docs'
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: 'actions/checkout@v4'
+
+      - name: 'Validate URLS into PhpDoc'
+        uses: 'dagger/dagger-for-github@8.0.0'
+        with:
+          version: "latest"
+          args: "test validate-url-doc"
```
Jean-Beru:
```suggestion
          version: 'latest'
          call: 'test validate-url-doc'
          dagger-flags: '--silent'
```

StevenRenaux: thanks

.github/workflows/url-test.yml
StevenRenaux: I don't get it?

.dagger/src/TestsGotenbergBundle.php
```diff
@@ -132,6 +142,11 @@ public function all(): array
             $this->phpstan(),
         ]);
 
+        $outputs[] = async(fn (): array => [
```
Neirda24: since it does not vary per php version should we add it to the matrix like this ?

.github/workflows/url-test.yml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+
+  check:
+    name: 'Validate URL docs'
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: 'actions/checkout@v4'
+
+      - name: 'Validate URLS into PhpDoc'
+        uses: 'dagger/dagger-for-github@8.0.0'
```
Neirda24, line 20:
```suggestion
        uses: 'dagger/dagger-for-github@v8.2.0'
```

composer.json
```diff
@@ -36,6 +36,7 @@
     },
     "require-dev": {
         "ext-mbstring": "*",
+        "amphp/http-client": "^5.3",
```
Neirda24: what for ?

Neirda24 (approved):
could we have a sample in the PR description on how the output looks like in case of failure please ?

composer.json
StevenRenaux: just issued about Multiplexing Responses for HttpClient doesn't work without this package. Don't understand why.

.github/workflows/url-test.yml
StevenRenaux: All other uses across the app are with `v8.0.0`

Neirda24: my bad. incoming PR with updates :)

composer.json
Neirda24: hmm let's discuss this together I think

StevenRenaux: remove Multiplexing Responses

Jean-Beru:
Thanks @StevenRenaux  !

PR #186, opened 2025-09-16
Add configurator on BuilderInterface

StevenRenaux:
| Bug fix ?               | yes/no   |

Add Configurator reference to BuilderInterface::class

### DIFF
--- a/src/DependencyInjection/CompilerPass/GotenbergPass.php
+++ b/src/DependencyInjection/CompilerPass/GotenbergPass.php
@@ -28,8 +28,12 @@ final class GotenbergPass implements CompilerPassInterface
             ;
 
             $class = $serviceDefinition->getClass();
+            $builders = $this->builderStack->getBuilders();
+            if (null === $class || !\array_key_exists($class, $builders)) {
+                throw new \RuntimeException(\sprintf('Unknown builder class "%s"', $class));
+            }
 
-            $type = $this->builderStack->getBuilders()[$class];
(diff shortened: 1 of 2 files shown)


src/DependencyInjection/CompilerPass/GotenbergPass.php
```diff
@@ -28,8 +28,12 @@ public function process(ContainerBuilder $container): void
             ;
 
             $class = $serviceDefinition->getClass();
+            $builders = $this->builderStack->getBuilders();
+            if (null === $class || !\array_key_exists($class, $builders)) {
+                throw new \RuntimeException(\sprintf('Unknown builder class "%s"', $class));
```
Jean-Beru: Error message will be weird if `$class` is null.

Jean-Beru: Can you give us an example where `$builders[$class]` does not exist ? Maybe we could just ignore the builder instead of throwing an exception.

StevenRenaux: In our case, I don't think we can generate this error. I'll type `$class` directly

StevenRenaux: You can see the next commit @Jean-Beru

Jean-Beru (approved):

Neirda24 (approved):

Neirda24:
thanks @StevenRenaux

PR #187, opened 2025-09-18
feat: Forward Authentication Cookie

### LINKED ISSUE #35: Forward cookie from request
Add a `forwardCookie(string $name)` that would fetch a cookie from the request and send it to Gotenberg.

It could be useful to authenticate Gotenberg when it tries to access to an URL.

_Originally posted by @Neirda24 in https://github.com/sensiolabs/GotenbergBundle/pull/33#discussion_r1565828062_

StevenRenaux:
| Issues                  | Fix #35  |


src/Builder/Behaviors/Dependencies/RequestContextAwareTrait.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Sensiolabs\GotenbergBundle\Builder\Behaviors\Dependencies;
+
+use Symfony\Component\DependencyInjection\Attribute\Autowire;
+use Symfony\Component\Routing\RequestContext;
+use Symfony\Contracts\Service\Attribute\SubscribedService;
+use Symfony\Contracts\Service\ServiceSubscriberTrait;
+
+trait RequestContextAwareTrait
```
StevenRenaux, line 10: Need to have the mofication from https://github.com/sensiolabs/GotenbergBundle/pull/188
A rebased will be necessary

composer.json
```diff
@@ -32,6 +32,7 @@
         "symfony/http-foundation": "^6.4 || ^7.0",
         "symfony/http-kernel": "^6.4 || ^7.0",
         "symfony/mime": "^6.4 || ^7.0",
+        "symfony/security-core": "^6.4 || ^7.0",
```
Jean-Beru: IMO, it should not be required to generate PDFs. What about marking it optional like Twig ?

src/Builder/Behaviors/Chromium/CookieTrait.php
```diff
@@ -102,6 +108,52 @@ public function forwardCookie(string $name): static
...
+    {
+        $request = $this->getRequestStack()->getCurrentRequest();
+
+        if (null === $request) {
+            $this->getLogger()?->debug('Cookie cannot be forwarded with authentication because there is no Request.');
+
+            return $this;
+        }
+
+        $request->getSession()->save();
```
Jean-Beru, line 124: I'm not sure of the impact of this call. Saving session could lead to unwanted behaviors since it's usually done at the end of the request.

src/Builder/Behaviors/Chromium/CookieTrait.php
```diff
@@ -102,6 +108,52 @@ public function forwardCookie(string $name): static
...
+    /**
+     * For CLI generation usage.
+     */
+    public function asUser(UserInterface $user, string $firewallName = 'main'): static
+    {
+        if (!class_exists(UsernamePasswordToken::class)) {
+            throw new \LogicException(\sprintf('UsernamePasswordToken is required to use "%s" method. Try to run "composer require symfony/security-bundle".', __METHOD__));
+        }
+
+        $token = new UsernamePasswordToken($user, $firewallName, $user->getRoles());
```
Jean-Beru, line 134: We should use an Authenticator since it does additional steps on login (like [user checkers](https://symfony.com/doc/current/security/user_checkers.html)).

The SecurityBundle already provides a method for allowing interactive login. If we want to only require the component, not the bundle, we could take a look to `AuthenticatorManager::authenticateUser` :thinking:

src/Builder/Behaviors/Chromium/CookieTrait.php
StevenRenaux: If we not save and close the session, we have a time out. Because Gotenberg call back our app to generate from a route or template and if the session is locked we have this time out

composer.json
StevenRenaux: I wanted to to that but because of the typing Userinterface into `asUser` method (https://github.com/StevenRenaux/GotenbergBundle/blob/features/Forward-auth-cookies/src/Builder/Behaviors/Chromium/CookieTrait.php#L132) I need to get it actually

PR #192, opened 2025-10-22
feat: Add flatten field for HtmlPdfBuilder - MarkdownPdfBuilder - UrlPdfBuilder - ConvertPdfBuilder

StevenRenaux:
| Q                       | A        |
|-------------------------|----------|
| Gotenberg API version ? | 8.16.0  |
| Bug fix ?               | no   |
| New feature ?           | yes   |
| BC break ?              | no   |
| Issues                  | Fix #191  |

## Description

Add flatten for `HtmlPdfBuilder` - `MarkdownPdfBuilder` - `UrlPdfBuilder` and  `ConvertPdfBuilder`

PR #193, opened 2025-10-22
Update doc generation

StevenRenaux:

## Description

Related to the last PR #192 
I realised how some documentation are missing. Rather than copy paste in some builders some methods, I think it would be relevant to follow the same structure as traits to avoid missing methods. It also explain by the same time the traits behavior.


Jean-Beru:
Usage example can be grabbed from `@example` annotation. It will allow to generate `docs/{pdf,screenshot}/*.md` files too.

StevenRenaux:
Next step:

Rather than generate all builders documentation from scratch, it might be more relevant to add only the undocumented methods in the markdown,  to be more precise for some of them.

StevenRenaux:
All builders documentation is generated from `generate.php` file
The doc generation is done under `<!-- AUTO-GENERATED:START -->` comment in all builders documentation files.

docs/generate.php
```diff
@@ -104,11 +86,19 @@ class BuilderParser
         'getSubscribedServices',
     ];
 
-    private string $name;
+    private const EXAMPLE_TEMPLATE = '
+```php
+return $gotenberg
+    // Your builder call as ->html() and the rest of your configuration code
+    {% method %}
```
Jean-Beru:
```suggestion
    ->{% method %}
```

to define an example without prefixing it with `->` . Simpler IMO.

docs/generate.php
```diff
@@ -162,10 +158,33 @@ public function extract(): string
...
+                $markdown = str_replace('{% method %}', $example, self::EXAMPLE_TEMPLATE);
+
+                $isLast = $lastKey === $key;
+
+                if (false === $isLast) {
+                    $markdown .= '<br />';
+                }
+            }
+
+            return rtrim($markdown, '<br />');
```
Jean-Beru: The second `rtrim` argument a list of unwanted characters, not a "full string".

docs/generate.php
```diff
@@ -162,10 +158,33 @@ public function extract(): string
...
+
+            foreach ($exampleList as $key => $example) {
+                $markdown = str_replace('{% method %}', $example, self::EXAMPLE_TEMPLATE);
+
+                $isLast = $lastKey === $key;
+
+                if (false === $isLast) {
+                    $markdown .= '<br />';
+                }
+            }
```
Jean-Beru:
```suggestion
            $lastKey = array_key_last($exampleList);
            $markdown = '';
            foreach ($exampleList as $key => $example) {
                $markdown .= str_replace('{% method %}', $example, self::EXAMPLE_TEMPLATE);
                if ($key !== $lastKey) {
                    $markdown .= '<br />';
                }
            }
```

docs/generate.php
```diff
@@ -182,14 +201,14 @@ public function extract(): string
                 $markdown .= $see."\n";
             }
 
+            $example = $renderExample($parts['tags']['example'] ?? []);
+            if ('' !== $example) {
+                $markdown .= $example;
+            }
+
             return $markdown;
         };
```
Jean-Beru: What about defining a function per part? Something like `renderDescription`, `renderlinks` and `renderExamples` ?

Then, the final content could be generated using
```php
$markdown = $renderDescription($parts) . $renderlinks($parts) . $renderExamples($parts);
```

IMO, splitting markdown generation by distinct parts is much easier to understand.

docs/screenshot/UrlScreenshotBuilder.md
```diff
@@ -0,0 +1,661 @@
...
+            ->route('home', [
+                'my_var' => 'value'
+            ])
+            ->generate()
+            ->stream()
+        ;
+    }
+}
+```
+<!-- AUTO generated doc from generate.php -->
```
Jean-Beru, line 63: Is this comment needed?

Jean-Beru:
Much clearer IMO since it limit "clicks" :tada: 

Before

https://github.com/StevenRenaux/GotenbergBundle/blob/1.x/docs/pdf/html-builder.md

After

https://github.com/StevenRenaux/GotenbergBundle/blob/features/Update-docs/docs/pdf/HtmlPdfBuilder.md

docs/generate.php
```diff
@@ -162,10 +158,33 @@ public function extract(): string
...
+         */
+        $renderExample = static function (array $exampleList): string {
+            if ([] === $exampleList) {
+                return '';
+            }
+
+            $lastKey = array_key_last($exampleList);
+
+            foreach ($exampleList as $key => $example) {
+                $markdown = str_replace('{% method %}', $example, self::EXAMPLE_TEMPLATE);
```
Neirda24: maybe leverage https://www.php.net/manual/fr/function.strtr.php ?

Neirda24 (commented):
really impressive work ! congrats. does methods still "grouped" by tag to ensure coherent ordering of methods ?

docs/screenshot/UrlScreenshotBuilder.md
StevenRenaux: Totally necessary for the generation :wink: 
You can write some specific content for a builder above and under it's auto generated

docs/generate.php
Jean-Beru: Or just use [heredoc](https://www.php.net/manual/fr/language.types.string.php#language.types.string.syntax.heredoc) ?

StevenRenaux: heredoc ;)

docs/screenshot/UrlScreenshotBuilder.md
Jean-Beru: The 2 comments are needed? 1 can be enough, right?

StevenRenaux: Just for information, and to prevent from anyone who might add content below without paying attention.

StevenRenaux:
> really impressive work ! congrats. does methods still "grouped" by tag to ensure coherent ordering of methods ?

@Neirda24 
WDYM? I don't modified the order I think

Jean-Beru (approved):

Jean-Beru:
Thanks @StevenRenaux :star_struck: !

PR #194, opened 2025-10-23
feat: Add support for versioning asset

### LINKED ISSUE #181: [Twig] Add better support for manifest.json file.
### Discussed in https://github.com/sensiolabs/GotenbergBundle/discussions/146

<div type='discussions-op-text'>

<sup>Originally posted by **thafirstone** February  4, 2025</sup>
Hi there,
I use `Webpack Encore` and so my templates contain versioned assets. When I use the `gotenberg_asset` twig function provided in the bundle, the paths are not well defined.
So I decided to create my own twig function as follows:
```
public function getCustomGotenbergAsset(string $path, bool $hasVersioning = true): string
{
    if ($hasVersioning) {
        $path = ltrim($this->container->get(Packages::class)->getUrl($path), '/');
    }

    return $this->container->get(Environment::class)->getRuntime(GotenbergAssetRuntime::class)->getAssetUrl($path);
}
```
It works very well. The only problem is that the `GotenbergAssetRuntime` class is marked `@Internal`, which means that it should not be called outside the bundle.
Is there another way to do this?</div>

StevenRenaux:
| Gotenberg API version ? | 8.23.1      |
| Issues                  | Fix #181  |

## Description

WDYT @Neirda24 ?


src/Builder/Behaviors/Chromium/AssetTrait.php
```diff
@@ -47,7 +47,12 @@ public function addAsset(string|\Stringable $path): static
             return $this;
         }
 
-        $assets[$path] = new \SplFileInfo($this->getAssetBaseDirFormatter()->resolve($path));
+        $splInfo = new \SplFileInfo($this->getAssetBaseDirFormatter()->resolve($path));
+        if (!$splInfo->isFile()) {
+            $splInfo = new \SplFileInfo($this->getAssetBaseDirFormatter()->resolve($path, true));
+        }
```
Jean-Beru: Should be in `AssetBaseDirFormatter`.

tests/Formatter/AssetBaseDirFormatterTest.php
```diff
@@ -10,23 +10,29 @@
...
+        yield 'absolute path and relative base dir' => ['/mock/foo/logo.png', '/bar', false, '/mock/foo/logo.png'];
+        yield 'relative path and relative base dir' => ['document.odt', 'bar/baz', false, '/mock/bar/baz/document.odt'];
+        yield 'relative path and absolute base dir' => ['foo/document.odt', '/mock/bar/baz', false, '/mock/bar/baz/foo/document.odt'];
+        yield 'relative path and relative base dir with end slash' => ['document.odt', 'bar/baz/', false, '/mock/bar/baz/document.odt'];
+
+        yield 'versioned absolute path and absolute base dir' => ['/mock/foo/file.md', '/mock/bar', true, '/mock/foo/file.md'];
+        yield 'versioned absolute path and relative base dir' => ['/mock/foo/logo.png', '/bar', true, '/mock/foo/logo.png'];
+        yield 'versioned relative path and relative base dir' => ['document.odt', 'bar/baz', true, '/mock/public/document.odt'];
+        yield 'versioned relative path and absolute base dir' => ['foo/document.odt', '/mock/bar/baz', true, '/mock/public/foo/document.odt'];
+        yield 'versioned relative path and relative base dir with end slash' => ['document.odt', 'bar/baz/', true, '/mock/public/document.odt'];
```
Jean-Beru: Should also test https://symfony.com/doc/current/components/asset.html#absolute-assets-and-cdns.

src/Formatter/AssetBaseDirFormatter.php
```diff
@@ -2,36 +2,49 @@
...
+            if (Path::isAbsolute($baseDir)) {
+                $filename = Path::join($baseDir, $path);
+                if (!file_exists($filename)) {
+                    continue;
+                }
+
+                return $filename;
+            }
+
+            $filename = Path::join($this->projectDir, $baseDir, $path);
```
Jean-Beru: Maybe we can deprecate this usage (`assets` instead of `%kernel.project_dir%/`) in another PR ?

Dev side: the configuration value will be easier to understand.
Code side: we won't have to inject `%kernel.project_dir%`.

WDYT ?

src/DependencyInjection/Configuration.php
```diff
@@ -24,9 +24,20 @@ public function getConfigTreeBuilder(): TreeBuilder
...
+                    ->prototype('scalar')->end()
                     ->info('Base directory will be used for assets, files, markdown')
-                    ->defaultValue('%kernel.project_dir%/assets')
+//                        ->children()
+//                            ->scalarNode('base_dir')
+//                                ->isRequired()
+//                                ->cannotBeEmpty()
+//                                ->defaultValue('%kernel.project_dir%/assets')
+//                            ->end()
+//                        ->end()
```
Jean-Beru: Should we keep these comments ?

tests/Formatter/AssetBaseDirFormatterTest.php
```diff
@@ -9,30 +9,27 @@
...
-        yield 'versioned relative path and relative base dir' => ['document.odt', 'bar/baz', true, '/mock/public/document.odt'];
-        yield 'versioned relative path and absolute base dir' => ['foo/document.odt', '/mock/bar/baz', true, '/mock/public/foo/document.odt'];
-        yield 'versioned relative path and relative base dir with end slash' => ['document.odt', 'bar/baz/', true, '/mock/public/document.odt'];
+        yield 'absolute path and absolute base dir' => [self::ASSETS_DIR.'/file.md', [self::ASSETS_DIR], self::ASSETS_DIR.'/file.md'];
+        yield 'absolute path and relative base dir' => [self::ASSETS_DIR.'/file.md', ['assets'], self::ASSETS_DIR.'/file.md'];
+        yield 'relative path and relative base dir' => ['file.md', ['Fixtures/assets'], \dirname(self::PROJECT_DIR, 2).'/Fixtures/assets/file.md'];
+        yield 'relative path and absolute base dir' => ['office/document.odt', [self::ASSETS_DIR], \dirname(self::PROJECT_DIR, 2).'/Fixtures/assets/office/document.odt'];
+        yield 'relative path and relative base dir with end slash' => ['document.odt', ['Fixtures/assets/office/'], \dirname(self::PROJECT_DIR, 2).'/Fixtures/assets/office/document.odt'];
```
Jean-Beru: We should also test multiple `baseDirectories`.

config/services.php
```diff
@@ -43,7 +43,9 @@
         ->tag('twig.extension')
     ;
     $services->set('sensiolabs_gotenberg.twig.asset_runtime', GotenbergRuntime::class)
+        ->call('setContainer')
         ->tag('twig.runtime')
+        ->tag('container.service_subscriber')
```
Jean-Beru: Why not just injecting `assets.packages` here? (with `NULL_ON_INVALID`)

src/Formatter/AssetBaseDirFormatter.php
StevenRenaux: I'm not totally sure because by default it's %kernel.project_dir%/assets which I think corresponds to several projects. Without this value each project will need to re-define the location of these assets

Jean-Beru: > I'm not totally sure because by default it's %kernel.project_dir%/assets

This default value will be absolute so this line won't be executed and no deprecation will be triggered.

I was talking about throw a deprecation when the value is other than an absolute path.

StevenRenaux: Ah, that's right, I didn't realize. That could be a solution and easier to maintain.

src/Formatter/AssetBaseDirFormatter.php
```diff
@@ -2,32 +2,53 @@
...
+     */
     public function __construct(
         private readonly string $projectDir,
-        string $baseDir,
+        array $baseDir,
     ) {
-        $this->baseDir = rtrim($baseDir, '/\\');
+        $this->baseDir = array_map(function ($value) {
+            return rtrim($value, '/\\');
+        }, $baseDir);
```
Jean-Beru:
```suggestion
        $this->baseDir = array_map(static fn ($value) => rtrim($value, '/\\'), $baseDir);
```

src/Formatter/AssetBaseDirFormatter.php
```diff
@@ -2,32 +2,51 @@
...
+            $filename = Path::join($this->projectDir, $baseDir, $path);
+            if (!file_exists($filename)) {
+                continue;
+            }
+
+            return $filename;
         }
 
-        return Path::join($this->projectDir, $this->baseDir, $path);
+        throw new FileNotFoundException(\sprintf('File "%s" does not exist.', $path));
```
Jean-Beru:
```suggestion
        throw new FileNotFoundException(\sprintf('File "%s" not found in assets directories: "%s".', $path, implode('", "', $this->baseDir)));
```

Jean-Beru (approved):
A last comment but 👌

src/Formatter/AssetBaseDirFormatter.php
```diff
@@ -2,32 +2,51 @@
...
-        if (Path::isAbsolute($this->baseDir)) {
-            return Path::join($this->baseDir, $path);
+        foreach ($this->baseDir as $baseDir) {
+            if (Path::isAbsolute($baseDir)) {
+                $filename = Path::join($baseDir, $path);
+                if (!file_exists($filename)) {
+                    continue;
+                }
+
+                return $filename;
```
Neirda24: might be worth "caching" it WDYT ?

tests/Builder/Pdf/FlattenPdfBuilderTest.php
```diff
@@ -54,7 +56,7 @@ public function testFilesExtensionRequirement(): void
         $this->expectExceptionMessage('The file extension "png" is not valid in this context.');
 
         $this->getBuilder()
-            ->files('b.png')
+            ->files(self::ASSETS_DIR.'logo.png')
```
Neirda24:
```suggestion
            ->files(self::ASSETS_DIR.'/logo.png')
```

composer.json
```diff
@@ -45,6 +46,7 @@
         "phpstan/phpstan-symfony": "^2.0",
         "phpunit/phpunit": "^10.5.46",
         "shipmonk/composer-dependency-analyser": "^1.8",
+        "symfony/asset": "^6.4 || ^7.0",
```
Neirda24:
```suggestion
        "symfony/asset": "^6.4 || ^7.0 || ^8.0",
```

src/Formatter/AssetBaseDirFormatter.php
```diff
@@ -2,32 +2,58 @@
...
+
+                return $this->resolvedPathsCache[$path] = $filename;
+            }
+
+            $filename = Path::join($this->projectDir, $baseDir, $path);
+            if (!file_exists($filename)) {
+                continue;
+            }
+
+            return $this->resolvedPathsCache[$path] = $filename;
```
Neirda24, line 49:
```suggestion
        foreach ($this->baseDir as $baseDir) {
            if (!Path::isAbsolute($baseDir)) {
                $baseDir = Path::join($this->projectDir, $baseDir);
            }

            $filename = Path::join($baseDir, $path);
            if (!file_exists($filename)) {
                continue;
            }

            return $this->resolvedPathsCache[$path] = $filename;
```

Something like this ?

tests/Builder/Pdf/FlattenPdfBuilderTest.php
```diff
@@ -22,6 +22,8 @@ final class FlattenPdfBuilderTest extends GotenbergBuilderTestCase
     /** @use WebhookTestCaseTrait<FlattenPdfBuilder> */
     use WebhookTestCaseTrait;
 
+    private const ASSETS_DIR = __DIR__.'/../../Fixtures/assets';
```
Neirda24: why does it need to be repated ?

src/Formatter/AssetBaseDirFormatter.php
StevenRenaux: done and thanks :wink:

src/Twig/GotenbergRuntime.php
```diff
@@ -14,6 +15,10 @@ final class GotenbergRuntime
 {
     private BuilderAssetInterface|null $builder = null;
 
+    public function __construct(private readonly Packages|null $packages)
```
Jean-Beru:
```suggestion
    public function __construct(private readonly Packages|null $packages =  null)
```
To  make it simpler to use (eg: avoid `new GotenbergRuntime(null)` in tests)

Neirda24 (approved):

src/Twig/GotenbergRuntime.php
StevenRenaux: Oups nice catch

Neirda24 (approved):

Jean-Beru:
Thanks @StevenRenaux for the PR and @thafirstone for the issue !

PR #198, opened 2025-11-02
Fully initialize the extension in `Bundle::createContainerExtension()` for compatibility with Symfony 7.4

GromNaN:
| Q                       | A        |
|-------------------------|----------|
| Gotenberg API version ? | 8.x      |
| Bug fix ?               | yes      |
| New feature ?           | no       |
| BC break ?              | yes      |
| Issues                  | Fix #197 |

## Description

In order to generate array shapes for all environments, Symfony FrameworkBundle needs to get the configuration of all the bundles without initializing the container.

The extension must be fully initialized in `Bundle::createContainerExtension()`.

PR #200, opened 2025-11-14
Fix Builder already added

StevenRenaux:

## Description

Since https://github.com/sensiolabs/GotenbergBundle/pull/198 we can't add a Custom Builder anymore in the Kernel::build method as 

```php
class Kernel extends BaseKernel
{
    use MicroKernelTrait;

    protected function build($container): void
    {
        /** @var SensiolabsGotenbergExtension $extension */
        $extension = $container->getExtension('sensiolabs_gotenberg');
        $extension->registerBuilder(InvoicePdfBuilder::class);
    }
}
```

During the first call you get an exception about the builder has already been added. And finally pass on the next request.
To fix that remove the exception and add return.

### DIFF
--- a/src/DependencyInjection/BuilderStack.php
+++ b/src/DependencyInjection/BuilderStack.php
@@ -46,7 +46,7 @@ final class BuilderStack
         }
 
         if (\array_key_exists($class, $this->builders)) {
-            throw new \LogicException("{$class} has already been added.");
+            return;
         }
 
         $reflection = new \ReflectionClass($class);


Neirda24 (approved):

Neirda24:
missing tests update maybe @StevenRenaux

StevenRenaux:
> missing tests update maybe @StevenRenaux

@Neirda24 We don't have test about BuilderStack. Tests failed because of the last CVE when we build the symfony container into dagger

Jean-Beru:
@StevenRenaux Do you know why the builder is registered twice? Maybe we can fix this instead.

Neirda24:
> @StevenRenaux Do you know why the builder is registered twice? Maybe we can fix this instead.

We investigated and I think it is because bundles have a mechanism that prevents extensions classes to be created twice (cached) but it doesn't apply to the build function of bundles or kernel.

All in all it actually makes sense to just ignore because the key is a class string. The chances two classes have same namespace and class name is really thin.

Jean-Beru:
> > @StevenRenaux Do you know why the builder is registered twice? Maybe we can fix this instead.
> 
> We investigated and I think it is because bundles have a mechanism that prevents extensions classes to be created twice (cached) but it doesn't apply to the build function of bundles or kernel.
> 
> All in all it actually makes sense to just ignore because the key is a class string. The chances two classes have same namespace and class name is really thin.

OK, LGTM!

Jean-Beru (approved):

Neirda24:
thank you @StevenRenaux

PR #199, opened 2025-11-10
fix(ci): Symfony dev version installation

Jean-Beru:
| Q                       | A        |
|-------------------------|----------|
| Gotenberg API version ? | 8.x      |
| Bug fix ?               | yes   |
| New feature ?           | no   |
| BC break ?              | no   |
| Issues                  |  |

## Description

Related to https://github.com/sensiolabs/GotenbergBundle/pull/198#pullrequestreview-3413264105.

PR #203, opened 2025-11-17
fix(ci): Symfony dev version installation

Jean-Beru:
| Issues                  | |

## Description

Replace #199.

Related to https://github.com/sensiolabs/GotenbergBundle/pull/198#pullrequestreview-3413264105.

This PR fixes CI by allowing `dev` packages when needed. It also add a functional test for the mentioned issue.

### DIFF
--- a/.dagger/src/GotenbergBundle.php
+++ b/.dagger/src/GotenbergBundle.php
@@ -75,11 +75,11 @@ class GotenbergBundle
             ->withMountedCache('/GotenbergBundle/vendor', $vendorCache)
             ->withExec(['composer', 'global', 'config', '--no-plugins', 'allow-plugins.symfony/flex', 'true'])
             ->withExec(['composer', 'global', 'require', 'symfony/flex'])
+            ->withExec(['composer', 'config', 'extra.symfony.require', $symfonyVersion])
         ;
         if (str_ends_with($symfonyVersion, '-dev')) {
             $phpContainer = $phpContainer
                 ->withExec(['composer', 'config', 'minimum-stability', 'dev'])
-                ->withExec(['composer', 'config', 'extra.symfony.require', $symfonyVersion])
             ;
         }
 
--- a/.dagger/src/GotenbergBundle.php
+++ b/.dagger/src/GotenbergBundle.php
@@ -15,7 +15,6 @@ use Dagger\Service;
 use function Amp\async;
 use function Amp\Future\await;
 use function Dagger\dag;
-use function sleep;
 
 #[DaggerObject]
 #[Doc('Module for GotenbergBundle')]
@@ -124,9 +123,10 @@ class GotenbergBundle
 
         string $phpVersion = '8.4',
         string $symfonyVersion = '7.3.*',
+        string $minimumStability = 'stable',
         Container|null $symfonyContainer = null,
     ): TestsGotenbergBundle {
-        $symfonyContainer ??= $this->symfonyContainer($source, $phpVersion, $symfonyVersion);
+        $symfonyContainer ??= $this->symfonyContainer($source, $phpVersion, $symfonyVersion, $minimumStability);
 
         return new TestsGotenbergBundle($symfonyContainer);
     }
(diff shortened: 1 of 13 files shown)


.dagger/src/GotenbergBundle.php
```diff
@@ -71,12 +71,20 @@ private function symfonyContainer(
...
             ->withMountedCache('/GotenbergBundle/vendor', $vendorCache)
-            ->withEnvVariable('SYMFONY_REQUIRE', $symfonyVersion)
             ->withExec(['composer', 'global', 'config', '--no-plugins', 'allow-plugins.symfony/flex', 'true'])
             ->withExec(['composer', 'global', 'require', 'symfony/flex'])
-            ->withExec(['composer', 'update'])
+        ;
+        if (str_ends_with($symfonyVersion, '-dev')) {
+            $phpContainer = $phpContainer
+                ->withExec(['composer', 'config', 'minimum-stability', 'dev'])
+                ->withExec(['composer', 'config', 'extra.symfony.require', $symfonyVersion])
```
Neirda24: IMO this should be outside (before) the if.

.dagger/src/TestsGotenbergBundle.php
```diff
@@ -52,10 +52,15 @@ public function validateDependencies(): string
 
     #[DaggerFunction]
     #[Doc('Run phpunit tests and returns the container it ran in.')]
-    public function phpunit(): string
+    public function phpunit(string $filter = ''): string
     {
+        $exec = ['./vendor/bin/phpunit', '--display-all-issues'];
+        if ($filter) {
```
Neirda24:
```suggestion
        if ('' !== $filter) {
```

.github/workflows/static.yml
```diff
@@ -41,7 +41,8 @@ jobs:
         uses: 'dagger/dagger-for-github@8.0.0'
         with:
           version: "latest"
-          args: "test php-cs-fixer"
+          call: "test php-cs-fixer"
+          dagger-flags: '--quiet'
```
Neirda24: ❤️

.dagger/src/GotenbergBundle.php
Jean-Beru: Good catch!

Neirda24 (approved):

.dagger/src/GotenbergBundle.php
```diff
@@ -69,14 +70,16 @@ private function symfonyContainer(
...
+        $composerCache = dag()->cacheVolume("php-{$phpVersion}-symfony-{$symfonyVersion}-composer-cache");
 
         return $phpContainer
-            ->withMountedCache('/GotenbergBundle/vendor', $vendorCache)
-            ->withEnvVariable('SYMFONY_REQUIRE', $symfonyVersion)
+            ->withMountedCache('/root/.composer/cache/files', $composerCache)
             ->withExec(['composer', 'global', 'config', '--no-plugins', 'allow-plugins.symfony/flex', 'true'])
             ->withExec(['composer', 'global', 'require', 'symfony/flex'])
-            ->withExec(['composer', 'update'])
+            ->withExec(['composer', 'config', 'audit.ignore', 'PKSA-365x-2zjk-pt47'])
```
Neirda24: Hmmm. I believe that is something that can be stored in the composer.json file ? Seems a bit out of place here. WDYT ?

dagger.json
```diff
@@ -1,12 +1,13 @@
...
   "sdk": {
     "source": "php"
   },
   "include": [
     "!.dagger/sdk",
     "!.dagger/vendor"
   ],
-  "source": ".dagger"
+  "source": ".dagger",
+  "disableDefaultFunctionCaching": true
```
Neirda24: It's dagger that automatically update this for backward compatibility. We should check if removing this setting still works and improve performance.

Jean-Beru: Good catch. I though I already reverted this unwanted change.

Neirda24: It is part of their new feature trying to not break everything. It is safer with this value to false. We need to make sure it doesn't have unwanted side effects if we remove it

Jean-Beru: My bad, I thought that I reverted this change.

StevenRenaux (approved):

Neirda24:
thank you @Jean-Beru

PR #204, opened 2025-11-17
fix(ci): remove matrix for PHPStan tests

Jean-Beru:
| Issues                  | |

## Description

Remove matrix for PHPStan tests since PR can't be merged when new templates are added for example. Ex: https://github.com/symfony/symfony/commit/3a65283953c0f233401f3afcdc1457638278bae7.

For now, CI is green due to a bug with next versions (ex: `7.4-dev` is ran with `7.3`). See https://github.com/sensiolabs/GotenbergBundle/pull/203.

### DIFF
--- a/.dagger/src/GotenbergBundle.php
+++ b/.dagger/src/GotenbergBundle.php
@@ -20,8 +20,12 @@ use function Dagger\dag;
 #[Doc('Module for GotenbergBundle')]
 class GotenbergBundle
 {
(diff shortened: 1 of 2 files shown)


Neirda24 (changes requested):
Not a fan of this... I think the matrix is important for static analysis too. Let's try another approach and see if it is hard to setup before merging this one.

Neirda24 (approved):

PR #207, opened 2025-11-20
Update dagger usage and versions

Neirda24:

## Description

Trying to improve dagger DX. First PR of several

### DIFF
--- a/.dagger/src/GotenbergBundle.php
+++ b/.dagger/src/GotenbergBundle.php
@@ -136,7 +136,7 @@ class GotenbergBundle
 
     #[DaggerFunction]
     #[Doc('Execute all tests within matrix (PHP version, Symfony version).')]
-    #[ReturnsListOfType('string')]
+    #[ReturnsListOfType(TestsGotenbergBundle::class)]
     public function testsMatrix(
         #[DefaultPath('.')]
         Directory $source,
@@ -144,7 +144,7 @@ class GotenbergBundle
         $tests = [];
 
         foreach ($this->getMatrix() as [$symfonyVersion, $phpVersion, $minimumStability]) {
-            $tests[] = async(fn () => $this->test($source, $phpVersion, $symfonyVersion, $minimumStability)->all());
+            $tests[] = async(fn () => $this->test($source, $phpVersion, $symfonyVersion, $minimumStability));
         }
 
         $result = [];
@@ -153,6 +153,6 @@ class GotenbergBundle
             $result[] = $test;
         }
 
-        return array_merge(...$result);
+        return $result;
     }
 }
--- a/.dagger/src/GotenbergBundle.php
+++ b/.dagger/src/GotenbergBundle.php
@@ -12,6 +12,7 @@ use Dagger\Attribute\ReturnsListOfType;
 use Dagger\Container;
 use Dagger\Directory;
 use Dagger\Service;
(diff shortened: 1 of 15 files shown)


.github/workflows/doc.yml
```diff
@@ -17,20 +17,19 @@ jobs:
...
-          php-version: '8.3'
-          coverage: 'none'
-
-      - name: 'Install dependencies'
-        uses: 'ramsey/composer-install@v3'
-
-      - name: 'Generate builder API doc'
-        run: 'php ./docs/generate.php'
+          version: "latest"
+          call: "generate-docs as-patch contents"
```
Jean-Beru: Single quotes?

.github/workflows/doc.yml
```diff
@@ -17,20 +17,19 @@ jobs:
...
+          dagger-flags: '--silent'
+        env:
+          DAGGER_NO_NAG: '1'
 
       - name: 'Check documentation'
         run: |
+          echo '${{ steps.generate-docs-patch.outputs.output }}' | git apply --allow-empty
           DIFF=$(git status --short --renames ./docs)
           if [[ $DIFF != '' ]]; then
             echo "::warning::Builders API documentation is not up to date. Please run \"php ./docs/generate.php\""
```
Jean-Beru: Run Dagger instead?

.dagger/src/GotenbergBundle.php
```diff
@@ -79,7 +81,7 @@ private function symfonyContainer(
             $minimumStability = 'stable';
         }
 
-        $composerCache = dag()->cacheVolume("php-{$phpVersion}-symfony-{$symfonyVersion}-composer-cache");
+        $composerCache = dag()->cacheVolume('composer-cache');
```
Jean-Beru: Is it better to have one cache with different Symfony versions than many smaller ones?

Neirda24: I believe so. Because in the end several libraries will be shared. also if we find a way (depends on dagger next features) to export that cache as an artifact to be shared accross jobs then I think it would be easier and better with a single cache. WDYT ? also having the `symfonyVersion` part of the key implies `7.3.*` is a different cache than `^7.3`

.dagger/src/Test/PhpCsFixer.php
```diff
@@ -0,0 +1,104 @@
...
+
+        $globalDataDir = trim($symfonyContainer->withExec(['composer', 'global', 'config', 'data-dir'])->stdout());
+        $globalBinDir = trim($symfonyContainer->withExec(['composer', 'global', 'config', 'bin-dir'])->stdout());
+
+        $this->symfonyContainer = $symfonyContainer
+            ->withEnvVariable(
+                'PATH',
+                "{$symfonyContainer->envVariable('PATH')}:{$globalDataDir}/{$globalBinDir}",
+            )
+        ;
```
Jean-Beru: It could be put in the Symfony's container build step because it's related to Composer. WDYT?

Neirda24: updated 👍

.dagger/src/GotenbergBundle.php
Jean-Beru: OK let's see that when we will store the composer cache as an artifact. :+1:

Jean-Beru (approved):

PR #214, opened 2025-12-04
fix: assert form DataPart name in assertGotenbergFormDataFile

Jean-Beru:
| BC break ?              | maybe   |
| Issues                  |  |

## Description

Fix assertion on form DataPart files by adding a test on its name.

:warning: Is it considered as a BC break since CI may fail after this fix. If so, we could do that before next major?
```php
if ($part->getContentType() === $contentType && $body instanceof File && $body->getPath() === Path::canonicalize($path)) {
    if ($part->getName() !== $name) {
        // warn
    }
    $this->addToAssertionCount(1);
}
```

### DIFF
--- a/src/Test/Builder/GotenbergBuilderTestCase.php
+++ b/src/Test/Builder/GotenbergBuilderTestCase.php
@@ -113,7 +113,7 @@ abstract class GotenbergBuilderTestCase extends TestCase
             }
 
             $body = \Closure::bind(static fn (DataPart $part) => $part->body, null, TextPart::class)($part);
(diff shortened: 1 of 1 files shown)


Neirda24:
IMO if it was the intended behavior it passes the BC layer

Jean-Beru:
> IMO if it was the intended behavior it passes the BC layer

I agree :slightly_smiling_face:

StevenRenaux (approved):

Neirda24 (approved):

PR #223, opened 2025-12-15
fix(version): Late collect version only if files were generated.

### LINKED ISSUE #221: bundle client hits /version on each request
hey, I dont know if its a misconfig on my side or else, since upgrading to v1.1 it appears the bundle tries to hit the server /version url on each request, even if I dont render/compute/stream any pdf on this said request

is it intentional ? 

I can provide other detail after further investigation, just wanted to issue this at first

thanks

### LINKED ISSUE #222: Allow %env(...)% in sensiolabs_gotenberg.version parameter
Hi,

Can you allow  %env(...)% for version in teh YAML configuration file ? We need it to properly start app when Gotenberg has not yet started.
Apparently setting the version control in DI/Configuration.php happens too early for env variables to work.

Thanks.

Neirda24:
| Issues                  | Fix #221, Fix #222 |


Neirda24:
🤝 merci @Neirda24

PR #224, opened 2025-12-15
chore: add upgrade file doc

94noni:
| Bug fix ?               | yesish (avoid http call in dev/profiler)   |
| Issues                  | Related to  https://github.com/sensiolabs/GotenbergBundle/issues/221#issuecomment-3656159240 |

## Description

i propose to add somewhere a doc for this "feature" 
friendly ping @Neirda24 do you think it is worth it ? in this file or directly in the readme ?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ know.
 * Create a new branch for your changes.
 * Ensure your code follows the existing style and includes tests if applicable.
 * Submit a pull request with a clear description of your changes.
-* Update the correct `UPGRADE-*.md` file if it introduce new deprecations, BC breaks or anything that may require changes on developers side.
+* Update the correct `UPGRADE-*.md` file if it introduces new deprecations, BC breaks or anything that may require changes on developers side.
 
 ## Installation
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,6 +18,7 @@ know.
 * Create a new branch for your changes.
 * Ensure your code follows the existing style and includes tests if applicable.
 * Submit a pull request with a clear description of your changes.
+* Update the correct `UPGRADE-*.md` file if it introduce new deprecations, BC breaks or anything that may require changes on developers side.
 
 ## Installation
 
--- a/UPGRADE-1.1.md
+++ b/UPGRADE-1.1.md
@@ -1,7 +1,7 @@
 # UPGRADE FROM 1.0.0 to 1.1.0
 
 ## Configuration Changes
-* Added new `sensiolabs_gotenberg.version` to add warning if trying to use a non-yet available feature from Gotenberg.
+* Added new `sensiolabs_gotenberg.version` to add warning when trying to use a feature not yet available in Gotenberg.
 
 ## Contributions changes
 * Move to [dagger.io](https://dagger.io/) to run tests both locally and in CI.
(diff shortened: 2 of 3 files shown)


Neirda24:
Hello @94noni . You are absolutely right we should have added this file. Although there are several lines to add from https://github.com/sensiolabs/GotenbergBundle/compare/v1.0.0...v1.1.0 just a few that comes to mind :
* `sensiolabs_gotenberg.version` as you mentionned
* New way to contribute using `Dagger.io`
* Auto `configurator` for service simplementing `BuilderInterface` and `autoconfigure: true`
* ...

Would you create this file or do you prefer we update this PR ?

94noni:
i do prefer you take over on this PR if possible (more knowledge of the internal) ;) 
but indeed having this documented somewhere is to me important to avoid such useless http queries

Jean-Beru (approved):
Good catch!

Jean-Beru:
Sorry, I didn't see your message before posting. 

> New way to contribute using Dagger.io
> Auto configurator for service simplementing BuilderInterface and autoconfigure: true

If so, updating this file must be mandatory for each PR that: adds a feature, deprecations, BC breaks, etc. (everything except doc and bugfix)

Neirda24:
@Jean-Beru : taking over this PR right now. I'll complete the upgrade-1.1 + create the upgrade-1.2 and update the Contributing guide as well.

CONTRIBUTING.md
```diff
@@ -18,6 +18,7 @@ know.
 * Create a new branch for your changes.
 * Ensure your code follows the existing style and includes tests if applicable.
 * Submit a pull request with a clear description of your changes.
+* Update the correct `UPGRADE-*.md` file if it introduce new deprecations, BC breaks or anything that may require changes on developers side.
```
smnandre:
```suggestion
* Update the correct `UPGRADE-*.md` file if it introduces new deprecations, BC breaks or anything that may require changes on developers side.
```

UPGRADE-1.1.md
```diff
@@ -0,0 +1,10 @@
+# UPGRADE FROM 1.0.0 to 1.1.0
+
+## Configuration Changes
+* Added new `sensiolabs_gotenberg.version` to add warning if trying to use a non-yet available feature from Gotenberg.
```
smnandre:
```suggestion
* Added new `sensiolabs_gotenberg.version` to add warning when trying to use a feature not yet available in Gotenberg.
```

Jean-Beru (approved):

UPGRADE-1.1.md
```diff
@@ -0,0 +1,10 @@
+# UPGRADE FROM 1.0.0 to 1.1.0
+
+## Configuration Changes
+* Added new `sensiolabs_gotenberg.version` to add warning when trying to use a feature not yet available in Gotenberg.
```
Jean-Beru, line 4: We could also specify that the version will be retrieved with a HTTP call if not set.

94noni: as it was kind the reason of my first issue I think its a good idea 👍🏻

Jean-Beru:
```suggestion
* Added new `sensiolabs_gotenberg.version` to add warning when trying to use a feature not yet available in Gotenberg.
  > [!NOTE]
  > If no version is explicitly defined in the configuration, the bundle will perform an additional HTTP call to the 
  > Gotenberg API to resolve the compatible version automatically.
```
?

Neirda24 (approved):

Neirda24:
thanks @94noni

PR #228, opened 2026-01-17
fix: Asset gotenberg runtime for asset mapper or webpack usage

### LINKED ISSUE #218: [Bug]: Could not render template due to missing CSS asset
#### Install Symfony 7.4
```
$ git clone https://github.com/dunglas/symfony-docker.git sf74
$ cd sf74/
$ docker compose build --pull --no-cache
$ SYMFONY_VERSION=7.4.* docker compose up --wait
```

#### Install Asset Mapper, Gotenberg bundle, Bootstrap
```
$ docker compose exec php composer require symfony/asset-mapper symfony/asset symfony/twig-pack
$ docker compose exec php composer require sensiolabs/gotenberg-bundle 1.1
$ docker compose exec php bin/console importmap:require bootstrap
```

#### Add Bootstrap in assets/app.js
```js
import 'bootstrap/dist/css/bootstrap.min.css';
```

#### Modify templates/base.html.twig
```twig
<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="utf-8">
        <meta name="viewport" content="width=device-width, initial-scale=1">
        {% block javascripts %}
            {% block importmap %}{{ importmap('app') }}{% endblock %}
        {% endblock %}
    </head>
    <body>
    <div class="col-lg-8 mx-auto p-4 py-md-5">
        <div class="card">
            <div class="card-body">
                <h5 class="card-title">Lorem Ipsum</h5>
                <p class="card-text">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor
                    incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco
                    laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate
                    velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt
                    in culpa qui officia deserunt mollit anim id est laborum.</p>
                <a href="{{ path('test_pdf') }}" target="_blank" class="card-link">Generate PDF</a>
            </div>
        </div>
    </body>
</html>
```

#### Add templates/doc.html.twig
```twig
<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="utf-8">
        <link href="{{ gotenberg_asset('vendor/bootstrap/dist/css/bootstrap.min.css') }}" rel="stylesheet"/>
    </head>
    <body>
    <div class="col-lg-8 mx-auto p-4 py-md-5">
        <div class="card">
            <div class="card-body">
                <h5 class="card-title">Lorem Ipsum</h5>
                <p class="card-text">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor
                    incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco
                    laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate
                    velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt
                    in culpa qui officia deserunt mollit anim id est laborum.</p>
            </div>
        </div>
    </body>
</html>
```

#### Create controller
```php
<?php

namespace App\Controller;

use Sensiolabs\GotenbergBundle\GotenbergPdfInterface;
use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
use Symfony\Component\HttpFoundation\Response;
use Symfony\Component\Routing\Attribute\Route;

class TestController extends AbstractController
{
    #[Route('/test', name: 'test')]
    public function test(): Response
    {
        return $this->render('base.html.twig');
    }

    #[Route('/test/pdf', name: 'test_pdf')]
    public function testPdf(GotenbergPdfInterface $gotenberg): Response
    {
        return $gotenberg->html()
            ->content('doc.html.twig')
            ->generate()
            ->stream();
    }
}
```

#### Rebuild and test it
```
$ docker compose up --build
```

=> https://localhost/test

The app crash:
> Could not render template "doc.html.twig" into PDF part "index.html". An exception has been thrown during the rendering of a template ("File "assets/vendor/bootstrap/dist/css/bootstrap.min-z0zQt-h.css" not found in assets directories: "/app/assets".") in doc.html.twig at line 5.

Now, if you downgrade the bundle to v1.0 (`docker compose exec php composer require sensiolabs/gotenberg-bundle 1.0`), you'll see doc rendering works as expected.

### Description

Within a Docker context and asset mapper `symfony/asset-mapper` used to manage Bootstrap files, the problem occurs at rendering during the execution of `gotenberg_asset` Twig function.

**Notes**

1. Rendering works as expected with v1.0.0 (bug started with v1.1.0)
2. Whereas the reproducer uses Bootstrap as CSS to demonstrate the bug, it should break as well with other assets

### Gotenberg Version

8.25.1

### Gotenberg Bundle Version

1.1.0

StevenRenaux:
| Issues                  | Fix #218  |

## Description

Tested with projets running with asset-mapper, webpack, and both when using webpack and symfonycasts/sass-bundle.

### DIFF
--- a/src/Twig/GotenbergRuntime.php
+++ b/src/Twig/GotenbergRuntime.php
@@ -4,6 +4,7 @@ namespace Sensiolabs\GotenbergBundle\Twig;
 
 use Sensiolabs\GotenbergBundle\Builder\BuilderAssetInterface;
 use Symfony\Component\Asset\Packages;
+use Symfony\Component\AssetMapper\AssetMapperRepository;
 
 /**
  * @internal
@@ -15,8 +16,10 @@ final class GotenbergRuntime
 {
     private BuilderAssetInterface|null $builder = null;
 
-    public function __construct(private readonly Packages|null $packages = null)
-    {
+    public function __construct(
+        private readonly Packages|null $packages = null,
+        private readonly AssetMapperRepository|null $assetMapperRepository = null,
+    ) {
     }
 
     public function setBuilder(BuilderAssetInterface|null $builder): void
@@ -73,6 +76,16 @@ final class GotenbergRuntime
 
     private function getVersionedPathIfExist(string $path): string
     {
+
+        $assetRepository = $this->assetMapperRepository;
+        if (null !== $assetRepository) {
+            $mappedPath = $assetRepository->find($path);
+
+            if (null !== $mappedPath) {
+                return $mappedPath;
+            }
+        }
+
         $packages = $this->packages;
         if (null !== $packages) {
             $path = ltrim($packages->getUrl($path), '/');
--- a/src/Twig/GotenbergRuntime.php
+++ b/src/Twig/GotenbergRuntime.php
@@ -76,7 +76,6 @@ final class GotenbergRuntime
 
     private function getVersionedPathIfExist(string $path): string
     {
-
         $assetRepository = $this->assetMapperRepository;
         if (null !== $assetRepository) {
             $mappedPath = $assetRepository->find($path);
--- a/tests/Twig/GotenbergRuntimeTest.php
+++ b/tests/Twig/GotenbergRuntimeTest.php
@@ -2,7 +2,6 @@
 
 namespace Sensiolabs\GotenbergBundle\Tests\Twig;
 
-use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\TestCase;
 use Sensiolabs\GotenbergBundle\Builder\BuilderAssetInterface;
 use Sensiolabs\GotenbergBundle\Twig\GotenbergRuntime;
@@ -69,21 +68,22 @@ class GotenbergRuntimeTest extends TestCase
         $builder = $this->createMock(BuilderAssetInterface::class);
         $builder->expects($this->once())
             ->method('addAsset')
-            ->with('foo')
+            ->with('/absolute/dir/vendor/package/dist/css/package.min.css')
         ;
 
         $assetMapperRepository = $this->createMock(AssetMapperRepository::class);
         $assetMapperRepository->expects($this->once())
             ->method('find')
-            ->willReturn('foo')
+            ->with('vendor/package/dist/css/package.min.css')
+            ->willReturn('/absolute/dir/vendor/package/dist/css/package.min.css')
         ;
 
         $runtime = new GotenbergRuntime(null, $assetMapperRepository);
         $runtime->setBuilder($builder);
 
-        $path = $runtime->getAssetUrl('foo');
+        $path = $runtime->getAssetUrl('vendor/package/dist/css/package.min.css');
 
-        $this->assertSame('foo', $path);
+        $this->assertSame('package.min.css', $path);
     }
 
     public function testGetAssetUrlWhenPackages(): void
@@ -91,21 +91,22 @@ class GotenbergRuntimeTest extends TestCase
         $builder = $this->createMock(BuilderAssetInterface::class);
         $builder->expects($this->once())
             ->method('addAsset')
-            ->with('foo')
+            ->with('image/example.png')
         ;
 
         $packages = $this->createMock(Packages::class);
         $packages->expects($this->once())
             ->method('getUrl')
-            ->willReturn('/foo')
+            ->with('image/example.png')
+            ->willReturn('/image/example.png')
         ;
 
         $runtime = new GotenbergRuntime($packages, null);
         $runtime->setBuilder($builder);
 
-        $path = $runtime->getAssetUrl('foo');
+        $path = $runtime->getAssetUrl('image/example.png');
 
-        $this->assertSame('foo', $path);
+        $this->assertSame('example.png', $path);
     }
 
     public function testGetAssetUrlWhenAssetMapperRepositoryAndPackages(): void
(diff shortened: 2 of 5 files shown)


src/Twig/GotenbergRuntime.php
```diff
@@ -73,6 +76,15 @@ private function addAsset(string $path, string $function): void
...
     private function getVersionedPathIfExist(string $path): string
     {
+        $assetRepository = $this->assetMapperRepository;
+        if (null !== $assetRepository) {
+            $mappedPath = $assetRepository->find($path);
+
+            if (null !== $mappedPath) {
+                return $mappedPath;
+            }
+        }
```
Jean-Beru, line 86: We should add some tests to ensure that a path is correctly resolved.

A dedicated documentation could also be added to to explain how Gotenberg handles assets (with or without Asset and AssetMapper components).

Neirda24 (approved):

src/Twig/GotenbergRuntime.php
StevenRenaux: Tests added

Regarding the documentation, I'm not sure how useful it will be. It's likely to raise more questions than it answers, since the resources should still be placed in the `asset` directory as usual.

tests/Twig/GotenbergRuntimeTest.php
```diff
@@ -60,4 +62,76 @@ public function testGetFontStyleTag(): void
...
+            ->willReturn(null)
+        ;
+
+        $runtime = new GotenbergRuntime($packages, $assetMapperRepository);
+        $runtime->setBuilder($builder);
+
+        $path = $runtime->getAssetUrl('foo');
+
+        $this->assertSame('foo', $path);
+    }
```
Jean-Beru, line 140: IMO, some combinations are missing, we should assert that:
- AssetMapper is used when installed
- Packages is used when installed and AssetMapper is missing
- Packages is used when installed and AssetMapper does not return a path
- Path is returned as it if neither AssetMapper nor Packages exists
- Path is returned as it if no corresponding asset is found

Maybe (not sure) a DataProvider could help here 🤔

src/Twig/GotenbergRuntime.php
Jean-Beru: Maybe we could just announce that the bundle is compatible with the Asset and AssetMapper components (thanks to you) .

dloubere:
@StevenRenaux Just to let you know that the fix works on my side. Thank you.

tests/Twig/GotenbergRuntimeTest.php
StevenRenaux: > IMO, some combinations are missing, we should assert that:
> 
> * AssetMapper is used when installed
> * Packages is used when installed and AssetMapper is missing
> * Packages is used when installed and AssetMapper does not return a path
> * Path is returned as it if neither AssetMapper nor Packages exists
> * Path is returned as it if no corresponding asset is found
> 
> Maybe (not sure) a DataProvider could help here 🤔

- AssetMapper is used when installed => `testGetAssetUrlWhenAssetMapperRepository`
- Packages is used when installed and AssetMapper is missing => `testGetAssetUrlWhenPackages`
- Packages is used when installed and AssetMapper does not return a path => `testGetAssetUrlWhenAssetMapperRepositoryAndPackages`
- Path is returned as it if neither AssetMapper nor Packages exists => `testGetAssetUrlWhenMissingAssetMapperRepositoryAndPackages`

- Path is returned as it if no corresponding asset is found => Already done by the others

tests/Twig/GotenbergRuntimeTest.php
```diff
@@ -60,4 +63,92 @@ public function testGetFontStyleTag(): void
...
+            ->method('find')
+            ->willReturn('foo')
+        ;
+
+        $runtime = new GotenbergRuntime(null, $assetMapperRepository);
+        $runtime->setBuilder($builder);
+
+        $path = $runtime->getAssetUrl('foo');
+
+        $this->assertSame('foo', $path);
```
Jean-Beru: Please use another value than `foo` since we can't know if the result comes from AssetMapper or the given path.

StevenRenaux: done ;)

src/Twig/GotenbergRuntime.php
Jean-Beru: Resolved, it is implicit.

Jean-Beru (approved):

src/Twig/GotenbergRuntime.php
smnandre: Are you sure this is working ? I mean.... AssetMapper is plugged into the Asset Packages, and the asset mapper repository si not used in prod so ... ?

Neirda24: Just tried it with APP_ENV=prod and it works even if the assets were not compiled. However It might be worth to mention @Jean-Beru , @StevenRenaux  that we do not support asset-mapper but mimic the way it work... @StevenRenaux : did you try with a `styles.scss` ? 

Because "importmap" cannot be supported out of the box. Since it dynamically access files on the same server but gotenberg expect us to send them every file.

StevenRenaux: I haven't tried with scss. Not sure if it works. I would be surprised if it works.
But I think the PDF needs to hold his own style. 
Maybe just mentioned that point into the documentation

PR #244, opened 2026-03-13
fix: Add normalize methods into dedicated traits

StevenRenaux:

## Description

When using `EmbedTrait` in a custom builder, I encountered an error with gotenberg.

After investigation, I found that the trait did not contain its `normalize` method as expected.

A fix was necessary wherever the `normalize` method was not implemented in the dedicated traits.
Some `files` methods were not being used by the dedicated trait.


Jean-Beru (approved):

adrienroches-sensio (approved):

Neirda24 (approved):

Neirda24:
thank you @StevenRenaux

PR #245, opened 2026-03-24
feat: Allow custom XMP metadata keys for Factur-X

Jean-Beru:
## Summary
- Allow arbitrary ExifTool-compatible metadata keys (e.g., `XMP-fx:DocumentType` for Factur-X) in both the `metadata()` PHP API and Symfony configuration
- Update `MetadataNodeBuilder` to use a variable prototype with `ignoreExtraKeys: false`, so custom keys pass through alongside the standard PDF metadata children
- Add tests for custom XMP metadata via `metadata()` and `addMetadata()`

## Test plan
- [x] `testSetCustomXmpMetadata` — verifies arbitrary XMP keys in `metadata()` are serialized correctly
- [x] `testAddCustomXmpMetadataToExisting` — verifies `addMetadata()` with custom keys merges properly
- [x] Full test suite passes (751 tests)
- [x] PHPStan passes with no errors

### DIFF
--- a/src/Builder/Behaviors/MetadataTrait.php
+++ b/src/Builder/Behaviors/MetadataTrait.php
@@ -21,22 +21,28 @@ trait MetadataTrait
      * @see https://gotenberg.dev/docs/convert-with-chromium/convert-html-to-pdf#metadata-pdf-engines
      * @see https://exiftool.org/TagNames/XMP.html#pdf
      *
-     * @param array{
-     *     Author?: string,
-     *     Copyright?: string,
-     *     CreationDate?: string,
-     *     Creator?: string,
-     *     Keywords?: string,
-     *     Marked?: bool,
-     *     ModDate?: string,
-     *     PDFVersion?: string,
-     *     Producer?: string,
-     *     Subject?: string,
-     *     Title?: string,
-     *     Trapped?: 'True'|'False'|'Unknown',
-     * } $metadata
+     * Common PDF metadata keys: Author, Copyright, CreationDate, Creator, Keywords,
+     * Marked, ModDate, PDFVersion, Producer, Subject, Title, Trapped.
      *
-     * @example metadata(['Author' => 'SensioLabs', 'Subject' => 'Gotenberg'])
+     * Any ExifTool-compatible key is accepted, including custom XMP namespaces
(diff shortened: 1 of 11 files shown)


src/Builder/Behaviors/MetadataTrait.php
```diff
@@ -21,22 +21,28 @@ abstract protected function getBodyBag(): BodyBag;
...
-     *     Trapped?: 'True'|'False'|'Unknown',
-     * } $metadata
+     * Common PDF metadata keys: Author, Copyright, CreationDate, Creator, Keywords,
+     * Marked, ModDate, PDFVersion, Producer, Subject, Title, Trapped.
      *
-     * @example metadata(['Author' => 'SensioLabs', 'Subject' => 'Gotenberg'])
+     * Any ExifTool-compatible key is accepted, including custom XMP namespaces
+     * (e.g., 'XMP-fx:DocumentType' for Factur-X).
+     *
+     * @param array<string, mixed>&array{
```
Jean-Beru, line 30: See https://phpstan.org/r/5d584f39-b519-47c1-83bf-541db3d3e69b

src/NodeBuilder/MetadataNodeBuilder.php
```diff
@@ -8,7 +8,6 @@ class MetadataNodeBuilder extends NodeBuilder implements NodeBuilderInterface
 {
     public function __construct(
         protected string $name,
-
```
Neirda24, line 11: is it really more readable ? Is it a cs-fixer rule ?

Neirda24 (approved):

src/NodeBuilder/MetadataNodeBuilder.php
Jean-Beru: I think so but it's just my personal opinion. I see 3 ways to declare this kind of argument:
- No lines between commented args
- 1 line between commented args
- Use @param at the function level
WDYT ?

Jean-Beru: I looked for a php-cs-fixer rule but nothing exists (it's just a stylistic formatter).

StevenRenaux (approved):

PR #246, opened 2026-03-27
fix: walk class hierarchy to discover NormalizeGotenbergPayload methods

Jean-Beru:
## Summary

- `ReflectionClass::getMethods()` on a child class does not return private methods from traits used in a parent abstract class (see https://3v4l.org/93481#v8.5.3)
- Instead of changing method visibility, the fix walks up the class hierarchy via `ReflectionClass::getParentClass()` to collect all `#[NormalizeGotenbergPayload]` methods from each level
- Added a unit test with dedicated fixture classes that reproduces the issue

### DIFF
--- a/src/Builder/AbstractBuilder.php
+++ b/src/Builder/AbstractBuilder.php
@@ -167,17 +167,23 @@ abstract class AbstractBuilder implements BuilderAsyncInterface, BuilderFileInte
         $normalizers = [];
 
         $reflection = new \ReflectionClass(static::class);
-        foreach (array_reverse($reflection->getMethods()) as $method) {
-            $attributes = $method->getAttributes(NormalizeGotenbergPayload::class);
+        do {
+            foreach (array_reverse($reflection->getMethods()) as $method) {
+                if ($method->getDeclaringClass()->getName() !== $reflection->getName()) {
+                    continue;
+                }
 
-            if (\count($attributes) === 0) {
-                continue;
-            }
+                $attributes = $method->getAttributes(NormalizeGotenbergPayload::class);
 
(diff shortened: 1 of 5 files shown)


Neirda24:
One other way would be to get the parent class from reflection to avoid leaking this kind of design. Protected might conflict each other on inheritance.

Neirda24 (approved):

Jean-Beru:
> One other way would be to get the parent class from reflection to avoid leaking this kind of design. Protected might conflict each other on inheritance.

Good catch! I made changes and added a test.

StevenRenaux:
> ## Summary
> * Changed all `#[NormalizeGotenbergPayload]` methods from `private` to `protected` visibility across all traits and builders
> * Private methods in traits are not accessible via `ReflectionMethod` when the trait is used in an abstract class hierarchy, which prevents payload normalization from working correctly
> 
> See https://3v4l.org/93481#v8.5.3 for a demonstration of the issue.

Still work if we extend a builder?

Jean-Beru:
> > ## Summary
> > 
> > * Changed all `#[NormalizeGotenbergPayload]` methods from `private` to `protected` visibility across all traits and builders
> > * Private methods in traits are not accessible via `ReflectionMethod` when the trait is used in an abstract class hierarchy, which prevents payload normalization from working correctly
> > 
> > See https://3v4l.org/93481#v8.5.3 for a demonstration of the issue.
> 
> Still work if we extend a builder?

Yep! `do { /* ... */ } while ($reflection = $reflection->getParentClass());` loop will parse all builder parents.

StevenRenaux (approved):

PR #249, opened 2026-04-06
feat: add watermark (release 8.28)

### LINKED ISSUE #269: [Feature] Add Stamp / Watermark operations
I am currently porting a DomPDF setup to Gotenberg and the existing solution prints an image on top of the PDF. Since Gotenberg is able to do this with the [stamp operation](https://gotenberg.dev/docs/manipulate-pdfs/stamp-pdfs), it would be great to have that functionality available in this bundle. I naively assume that watermarking is very similar in concept, that's why I included it with the same feature request.

StevenRenaux:
| Issues              | Fix #269   |

## Description
                                                                                                                                                                                                                                    
  - Add WatermarkSource enum (text, image, pdf)                                                                                                                                                                                     
  - Add WatermarkTrait with 5 fields: watermarkSource, watermarkExpression, watermarkPages, watermarkOptions, watermarkFile                                                                                                         
  - Add WatermarkPdfBuilder (POST /forms/pdfengines/watermark) with full payload validation (requires Gotenberg >= 8.28)                                                                                                            
  - Compose WatermarkTrait into ChromiumPdfTrait, LibreOfficeTrait, MergePdfBuilder, SplitPdfBuilder                                                                                                                                
  - Expose watermark() on GotenbergPdfInterface, GotenbergPdf, TraceableGotenbergPdf                                                                                                                                                
  - Add WatermarkTestCaseTrait + WatermarkPdfBuilderTest; extend existing builder tests                                                                                                                                             
  - Add 3 pre-existing NodeBuilder generics ignore entries to phpstan.dist.neon                                                                                                                                                     
  - Regenerate API docs


src/Builder/Pdf/WatermarkPdfBuilder.php
```diff
@@ -0,0 +1,79 @@
...
+ * @see https://gotenberg.dev/docs/manipulate-pdfs/watermark-pdfs
+ *
+ * @example files('document.pdf')
+ */
+#[WithBuilderConfiguration(type: 'pdf', name: 'watermark')]
+final class WatermarkPdfBuilder extends AbstractBuilder
+{
+    use AssetBaseDirFormatterAwareTrait;
+    use DownloadFromTrait;
+    use EmbedTrait;
```
Jean-Beru: Same issue as in #250: the `WatermarkPdfBuilder` includes `EmbedTrait`, `EncryptTrait`, `MetadataTrait`, and `PdfFormatTrait`, but the Gotenberg `/forms/pdfengines/watermark` route handler ([routes.go:1248-1297](https://github.com/gotenberg/gotenberg/blob/main/pkg/modules/pdfengines/routes.go#L1248-L1297)) only processes watermark-specific fields and the watermark file upload. It does **not** read pdfFormat, metadata, encryption, or embed form fields -- they would be silently ignored.

Composing `WatermarkTrait` into `MergePdfBuilder`/`SplitPdfBuilder` is correct since those routes do process all those fields.

_Review by Claude (claude.ai/code)_

Jean-Beru (commented):
Review by Claude (claude.ai/code).

This PR mirrors #250 (stamp) for watermark support. Same structure, same patterns. Three issues found -- two are shared with #250, one is specific to this PR.

src/Builder/Pdf/WatermarkPdfBuilder.php
```diff
@@ -0,0 +1,79 @@
...
+
+        if ($this->getBodyBag()->get('files') === null && $this->getBodyBag()->get('downloadFrom') === null) {
+            throw new MissingRequiredFieldException('At least one PDF file is required.');
+        }
+
+        if ($this->getBodyBag()->get('watermarkSource') === null) {
+            throw new MissingRequiredFieldException('Field "watermarkSource" must be provided.');
+        }
+
+        if ($this->getBodyBag()->get('watermarkExpression') === null) {
```
Jean-Beru, line 62: Same issue as in #250: `watermarkExpression` is validated as mandatory, but in the Gotenberg API it is parsed as optional (`form.String(prefix+"Expression", &expression, false)`).

For `image`/`pdf` sources, the expression is overwritten server-side by the uploaded file path, so it is effectively unused. The tests work around this by passing `->watermarkExpression('')`.

Suggestion: only require `watermarkExpression` when `watermarkSource` is `Text`.

_Review by Claude (claude.ai/code)_

Jean-Beru:
**Missing service registration and bundle wiring** -- _Review by Claude (claude.ai/code)_

Compared with #250 (stamp), this PR is missing three key files:

1. **`config/builder_pdf.php`** -- no service definition for `.sensiolabs_gotenberg.pdf_builder.watermark`. Without this, the DI container won't know about `WatermarkPdfBuilder`.

2. **`src/SensiolabsGotenbergBundle.php`** -- `WatermarkPdfBuilder::class` is not registered via `$extension->registerBuilder()`.

3. **`src/DependencyInjection/SensiolabsGotenbergExtension.php`** -- the `watermark` key is not added to the phpdoc type for the pdf config array.

Without these, calling `$gotenberg->watermark()` will fail at runtime because the builder won't be resolved from the container.

src/Builder/Pdf/WatermarkPdfBuilder.php
StevenRenaux: The Gotenberg API parses watermarkExpression as optional (form.String(..., false)), so technically valid for image/pdf sources without it.                                                                                      
                                                                                                                                                                                                                                  
  Agreed that the Go handler doesn't enforce it server-side. However, the https://gotenberg.dev/docs/manipulate-pdfs/watermark-pdfs marks watermarkExpression as required and defines its meaning for every source type:            
   
  - text → the string to render                                                                                                                                                                                                     
  - image / pdf → the filename of the uploaded watermark file                                                                                                                                                                     
                                                                                                                                                                                                                                    
  Since the field is documented as required and carries semantic meaning even for file-based sources, we chose to enforce it at the bundle level for all sources rather than diverge from the published API contract. This also     
  avoids silent, hard-to-debug omissions where a caller forgets the expression for an image/pdf watermark.

src/Builder/Behaviors/WatermarkTrait.php
```diff
@@ -0,0 +1,125 @@
...
+     * @see https://gotenberg.dev/docs/manipulate-pdfs/watermark-pdfs
+     *
+     * @example watermarkPages('1-3')
+     */
+    #[WithConfigurationNode(new ScalarNodeBuilder('watermark_pages'))]
+    public function watermarkPages(string $watermarkPages): self
+    {
+        $this->logWarningIfVersionIs('<', '8.28', 'The watermark option is not available.');
+
+        $this->getBodyBag()->set('watermarkPages', $watermarkPages);
```
Jean-Beru: Same comment as https://github.com/sensiolabs/GotenbergBundle/pull/250#discussion_r3045278053

src/Builder/Behaviors/WatermarkTrait.php
```diff
@@ -0,0 +1,125 @@
...
+
+        return $this;
+    }
+
+    #[NormalizeGotenbergPayload]
+    private function normalizeWatermark(): \Generator
+    {
+        yield 'watermarkSource' => NormalizerFactory::enum();
+        yield 'watermarkExpression' => NormalizerFactory::noop();
+        yield 'watermarkPages' => NormalizerFactory::noop();
```
Jean-Beru: Is it needed?

src/Builder/Pdf/WatermarkPdfBuilder.php
Jean-Beru: Let's follow documentation!

src/GotenbergPdfInterface.php
```diff
@@ -74,4 +75,9 @@ public function encrypt(): BuilderInterface;
      * @return EmbedPdfBuilder
      */
     public function embed(): BuilderInterface;
+
+    /**
+     * @return WatermarkPdfBuilder
+     */
+    public function watermark(): BuilderInterface;
```
Jean-Beru, line 82: Same as the PR about stamps. It's a BC break. Maybe we can skip this change and deprecate this interface if not useful.

StevenRenaux: Depending of what you decide with @Neirda24

Neirda24 (approved):

Jean-Beru (approved):

PR #250, opened 2026-04-06
feat: add Stamp (release 8.28)

### LINKED ISSUE #269: [Feature] Add Stamp / Watermark operations

StevenRenaux:
| Issues              | Fix #269   |

## Description

  Implements support for the new POST /forms/pdfengines/stamp route introduced in https://github.com/gotenberg/gotenberg/releases/tag/v8.28.0.                                                                                    
                                                                                                                                                                                                                                    
  - Adds StampPdfBuilder with StampTrait composing 5 fields: stampSource, stampExpression, stampPages, stampOptions, stampFile                                                                                                      
  - Adds StampSource enum (text | image | pdf)                                                                                                                                                                                      
  - Adds NormalizerFactory::stamp() for file upload normalization                                                                                                                                                                   
  - Composes StampTrait into all existing PDF builders: ChromiumPdfTrait, LibreOfficeTrait, MergePdfBuilder, SplitPdfBuilder                                                                                                        
  - Registers stamp() on GotenbergPdf, GotenbergPdfInterface, and TraceableGotenbergPdf                                                                                                                                             
  - Full test coverage: StampTestCaseTrait + StampPdfBuilderTest (happy paths for text/image/pdf sources, validation errors)                                                                                                        
  - Documentation generated via docs/generate.php


config/reference.php
```diff
@@ -0,0 +1,1632 @@
...
+    /**
+     * @param RoutesConfig $config
+     *
+     * @psalm-return RoutesConfig
+     */
+    public static function config(array $config): array
+    {
+        return $config;
+    }
+}
```
Jean-Beru: This file should be ignored

src/Builder/Behaviors/StampTrait.php
```diff
@@ -0,0 +1,127 @@
...
+
+        return $this;
+    }
+
+    #[NormalizeGotenbergPayload]
+    private function normalizeStamp(): \Generator
+    {
+        yield 'stampSource' => NormalizerFactory::enum();
+        yield 'stampExpression' => NormalizerFactory::noop();
+        yield 'stampPages' => NormalizerFactory::noop();
```
Jean-Beru: Is it needed?

src/Builder/Behaviors/StampTrait.php
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * Page ranges to stamp (e.g., '1-3', '5'). Empty string means all pages.
+     *
+     * @see https://gotenberg.dev/docs/manipulate-pdfs/stamp-pdfs
+     *
+     * @example stampPages('1-3')
+     */
+    #[WithConfigurationNode(new ScalarNodeBuilder('stamp_pages'))]
+    public function stampPages(string $stampPages): self
```
Jean-Beru: Could be nullable as `nativePageRanges` + `ValidatorFactory::range` to validate input.

src/Builder/Pdf/StampPdfBuilder.php
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @see https://gotenberg.dev/docs/manipulate-pdfs/stamp-pdfs
+ *
+ * @example files('document.pdf')
+ */
+#[WithBuilderConfiguration(type: 'pdf', name: 'stamp')]
+final class StampPdfBuilder extends AbstractBuilder
+{
+    use AssetBaseDirFormatterAwareTrait;
+    use DownloadFromTrait;
```
Jean-Beru, line 31: The `StampPdfBuilder` includes `EmbedTrait`, `EncryptTrait`, `MetadataTrait`, and `PdfFormatTrait`, but the Gotenberg `/forms/pdfengines/stamp` route handler ([routes.go:1299-1348](https://github.com/gotenberg/gotenberg/blob/main/pkg/modules/pdfengines/routes.go#L1299-L1348)) only processes stamp-specific fields (`stampSource`, `stampExpression`, `stampPages`, `stampOptions`) and the stamp file upload. It does **not** read pdfFormat, metadata, encryption, or embed form fields -- those would be silently ignored, which could mislead users.

By contrast, the merge and split routes do process all those fields, so composing `StampTrait` into `MergePdfBuilder`/`SplitPdfBuilder` is correct.

Suggestion: remove `EmbedTrait`, `EncryptTrait`, `MetadataTrait`, and `PdfFormatTrait` from `StampPdfBuilder`, unless there are plans to support them in the Gotenberg API.

Jean-Beru (commented):
Review by Claude (claude.ai/code).

Overall this is a clean PR that follows the existing builder patterns well. A few issues found below.

src/Builder/Pdf/StampPdfBuilder.php
```diff
@@ -0,0 +1,79 @@
...
+
+        if ($this->getBodyBag()->get('files') === null && $this->getBodyBag()->get('downloadFrom') === null) {
+            throw new MissingRequiredFieldException('At least one PDF file is required.');
+        }
+
+        if ($this->getBodyBag()->get('stampSource') === null) {
+            throw new MissingRequiredFieldException('Field "stampSource" must be provided.');
+        }
+
+        if ($this->getBodyBag()->get('stampExpression') === null) {
```
Jean-Beru, line 62: `stampExpression` is validated as mandatory here, but in the Gotenberg API it is parsed with `form.String(prefix+"Expression", &expression, false)` -- the `false` means it is **not** mandatory.

For `text` source, the expression is the text to render, so requiring it makes sense.
For `image`/`pdf` sources, the expression is overwritten server-side by the uploaded file path, so it is effectively unused.

The tests (`testHappyPathImageSource`, `testHappyPathPdfSource`) work around this by passing `->stampExpression('')`which feels like a workaround.

Suggestion: only require `stampExpression` when `stampSource` is `Text`.

_Review by Claude (claude.ai/code)_

StevenRenaux: The Gotenberg API parses stampExpression as optional (form.String(..., false)), so technically valid for image/pdf sources without it.

Agreed that the Go handler doesn't enforce it server-side. However, the https://gotenberg.dev/docs/manipulate-pdfs/stamp-pdfs marks stampExpression as required and defines its meaning for every source type:

text → the string to render
image / pdf → the filename of the uploaded watermark file
Since the field is documented as required and carries semantic meaning even for file-based sources, we chose to enforce it at the bundle level for all sources rather than diverge from the published API contract. This also
avoids silent, hard-to-debug omissions where a caller forgets the expression for an image/pdf watermark.

src/Builder/Pdf/StampPdfBuilder.php
StevenRenaux: Done

docs/pdf/HtmlPdfBuilder.md
```diff
@@ -294,6 +299,81 @@ return $gotenberg
...
+### stampPages(?string \$stampPages)
+Page ranges to stamp (e.g., '1-3', '5'). Empty string means all pages.<br />
+
+> [!TIP]
+> See: [https://gotenberg.dev/docs/manipulate-pdfs/stamp-pdfs](https://gotenberg.dev/docs/manipulate-pdfs/stamp-pdfs)
+
+```php
+return $gotenberg
+    // Your builder call as ->html() and the rest of your configuration code
+    ->stampPages('1-3')
```
Neirda24, line 377: Not possible to have one configuration for certain range and another for other ranges ?

src/GotenbergPdfInterface.php
```diff
@@ -74,4 +75,9 @@ public function encrypt(): BuilderInterface;
      * @return EmbedPdfBuilder
      */
     public function embed(): BuilderInterface;
+
+    /**
+     * @return StampPdfBuilder
+     */
+    public function stamp(): BuilderInterface;
```
Neirda24, line 83: Breaking change. WDYT @Jean-Beru ?

src/Builder/Pdf/StampPdfBuilder.php
Jean-Beru: OK, let's follow official documentation!

src/GotenbergPdfInterface.php
Jean-Beru: For sure 😎 

BTW do we really need this interface?

docs/pdf/HtmlPdfBuilder.md
Jean-Beru: I don't understand. Can you give an example? 🙏

Neirda24 (approved):

Jean-Beru (approved):

PR #252, opened 2026-04-09
feat: Allow to pass a custom callback on pdf stream process

kevin-verschaeve:
| Issues                  | - |

## Description

This allow the user to give a custom callback, to override the default stream process if needed.

To give an example, I need to save the generated PDF to an S3 before sending it to the client.
I am able to do it with this code change and this callback in my app:

```
->stream(function (ResponseStreamInterface $stream) use ($storagePath): void {
      $tempFile = tmpfile();

      foreach ($stream as $chunk) {
          $content = $chunk->getContent();
          echo $content;
          flush();
          fwrite($tempFile, $content);
      }

      rewind($tempFile);
      $this->storage->writeStream($storagePath, $tempFile);
      fclose($tempFile);
  })
```

Please let me know what you think of this :)

### DIFF
--- a/src/Builder/Result/GotenbergFileResult.php
+++ b/src/Builder/Result/GotenbergFileResult.php
@@ -106,7 +106,7 @@ class GotenbergFileResult extends AbstractGotenbergResult
         return $generator->getReturn();
     }
 
-    public function stream(): StreamedResponse
+    public function stream(callable $streamCallback = null): StreamedResponse
     {
         $this->ensureExecution();
         $this->processed = true;
@@ -119,19 +119,25 @@ class GotenbergFileResult extends AbstractGotenbergResult
             $headers['Content-Disposition'] = [HeaderUtils::makeDisposition($this->disposition, $filename)];
         }
 
+        $stream = $this->stream;
+
         return new StreamedResponse(
-            function () use ($filename): void {
-                if (!$this->stream->valid()) {
-                    throw new ProcessorException('Already processed query.');
+            $streamCallback
+                ? function () use ($streamCallback, $stream) {
+                    return $streamCallback($stream);
                 }
+                : function () use ($stream, $filename): void {
+                    if (!$stream->valid()) {
+                        throw new ProcessorException('Already processed query.');
+                    }
 
-                $generator = ($this->processor)($filename);
+                    $generator = ($this->processor)($filename);
 
-                foreach ($this->stream as $chunk) {
-                    $generator->send($chunk);
-                    echo $chunk->getContent();
-                    flush();
-                }
+                    foreach ($stream as $chunk) {
+                        $generator->send($chunk);
+                        echo $chunk->getContent();
+                        flush();
+                    }
             },
             $this->getStatusCode(),
             $headers,


Neirda24:
Hi. Thank you for your contribution. Is https://github.com/sensiolabs/GotenbergBundle/blob/1.x/docs/processing.md#sensiolabsgotenbergbundlebridgeasyncawsprocessorasyncawss3multipartprocessor not enough ? It does save to s3 and on top of that it can save and stream to teh client at the same time.

Jean-Beru:
> I need to save the generated PDF to an S3 before sending it to the client. 

The amazing @Neirda24 already wrote it for you https://github.com/sensiolabs/GotenbergBundle/blob/1.x/docs/processing.md#sensiolabsgotenbergbundlebridgeasyncawsprocessorasyncawss3multipartprocessor.

kevin-verschaeve:
oh !

I've tried with a TempFileProcessor, but I didn't liked it, because it was making me streaming the file twice, but maybe I used it wrong too.

I'll try with the AsyncAwsS3MultiPartProcessor and close this PR if everything is fine :)

kevin-verschaeve:
ok, I'm able to do it with the FlysystemProcessor, but I have to create a custom Processor over it to handle the file name for download and file path for storage. 

Not sure if I like it, but it works.

Maybe we can close this, I'll let you choose.

I'm putting here my workaround, to help future people if needed (or let you give some feedback)

```
  $storageProcessor = new FixedKeyProcessorDecorator(
    new FlysystemProcessor($this->storage),
    $storagePath,
);

return $pdf
    ->generate()
    ->processor($storageProcessor)
    ->stream();
````

```
final class FixedKeyProcessorDecorator implements ProcessorInterface
{
    /**
     * @param ProcessorInterface<T> $inner
     */
    public function __construct(
        private readonly ProcessorInterface $inner,
        private readonly string $key,
    ) {
    }

    public function __invoke(string|null $fileName): \Generator
    {
        return yield from ($this->inner)($this->key);
    }
}
```

**Comment by Jean-Beru** [comment]
Perhaps we could add a nullable string as constructor argument to override filename if set in filesystem-related processors like`FileProcessor`. Ex:
```php
final class FileProcessor implements ProcessorInterface
{
    public function __construct(
        private readonly Filesystem $filesystem,
        private readonly string $directory,
        private readonly LoggerInterface|null $logger = null,
        private readonly string|null $overrideFilename = null,
    ) {
    }
```
WDYT @Neirda24 ?

Neirda24:
@Jean-Beru : as discussed we need to define better what `->filename()` is about. Is it about the file that will be saved or the one that will be displayed / downloadable ?

One opinion would be to consider this :
The file being generated is from Gotenberg API. Thus we are just a client and the filename is required / provided by the API. So I think `->filename()` should be about the file to store. Meaning to help on @kevin-verschaeve issue we could leverage a `responseFilename()` or something like this that will override the filename when being downloaded. Customising per processor maybe. But then it would be something quite annoying to setup I think because we can have multiple processors. I need to dig a bit more on this

BTW @kevin-verschaeve can I ask why the use of `FlysystemProcessor` instead of AwsAsync ?

kevin-verschaeve:
I want to change the storage depending on the environment, working locally I use the filesystem to store the file.
I don't have a Local stack or équivalent, and didn't want to set it up just for this.
So flysystem seemed more legit :)

Jean-Beru:
> Customising per processor maybe. But then it would be something quite annoying to setup I think because we can have multiple processors. I need to dig a bit more on this

Having multiple processor can mean that each one will store the file in different ways. A `wither` can be a great alternative:
```php
$parts = [];
$parts[] = $builder
    /* ... */
    ->processor($processor->withFilename('1'))
    ->generate()
    ->process()
$parts[] = $builder
    /* ... */
    ->processor($processor->withFilename('2'))
    ->generate()
    ->process()

return $merger
    ->files(...$parts)
    ->generate()
    ->stream()
;
```

kevin-verschaeve:
Hello
I'm closing this PR as the proposed solution can be done differently.
Thanks for the help

PR #254, opened 2026-04-10
feat: add Rotate (release 8.28)

StevenRenaux:

## Description

● Implements support for the new POST /forms/pdfengines/rotate route introduced in https://github.com/gotenberg/gotenberg/releases/tag/v8.28.0.

  - Adds RotatePdfBuilder composing FilesTrait, DownloadFromTrait, and WebhookTrait — intentionally minimal to match the route's actual field support                                                                               
  - Adds RotateTrait exposing two fields: rotateAngle (required) and rotatePages (optional page range)
  - Adds RotateAngle enum (90 | 180 | 270) backed by int, normalized to string via NormalizerFactory::enum()                                                                                                                        
  - Registers rotate() on GotenbergPdf, GotenbergPdfInterface, and TraceableGotenbergPdf                                                                                                                                            
  - Full test coverage: RotateTestCaseTrait + RotatePdfBuilderTest (all three angles, page range, unsetting fields, missing file/angle validation errors)                                                                           
  - Documentation generated via docs/generate.php

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,3 @@
 .phpunit.result.cache
 composer.lock
 phpstan.neon
-/config/reference.php
--- a/docs/pdf/HtmlPdfBuilder.md
+++ b/docs/pdf/HtmlPdfBuilder.md
@@ -86,6 +86,8 @@ class YourController
 - [metadata](#metadataarray-metadata)
 - [pdfFormat](#pdfformatsensiolabsgotenbergbundleenumerationpdfformat-format)
 - [pdfUniversalAccess](#pdfuniversalaccessbool-bool)
+- [rotateAngle](#rotateanglesensiolabsgotenbergbundleenumerationrotateangle-rotateangle)
+- [rotatePages](#rotatepagesstring-rotatepages)
 - [splitMode](#splitmodesensiolabsgotenbergbundleenumerationsplitmode-splitmode)
 - [splitSpan](#splitspanstring-splitspan)
 - [splitUnify](#splitunifybool-bool)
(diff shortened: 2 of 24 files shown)


docs/pdf/HtmlPdfBuilder.md
```diff
@@ -246,6 +248,36 @@ return $gotenberg
...
+```php
+return $gotenberg
+    // Your builder call as ->html() and the rest of your configuration code
+    ->rotateAngle(RotateAngle::Rotate90)
+    ->generate()
+    ->stream()
+;
+```
+
+### rotatePages(?string \$rotatePages)
```
Neirda24, line 292: What happens if rotatePages without rotateAngle ? Also it means we can't rotate page 1 and 2 to 90 and page 3 to 180 ?

src/GotenbergPdfInterface.php
```diff
@@ -74,4 +75,9 @@ public function encrypt(): BuilderInterface;
      * @return EmbedPdfBuilder
      */
     public function embed(): BuilderInterface;
+
+    /**
+     * @return RotatePdfBuilder
+     */
+    public function rotate(): BuilderInterface;
```
Neirda24, line 94: This might be considered a BC change. @Jean-Beru WDYT we should do ? Maybe add a BC policy markdown file somewhere where we could list what is included / excluded.

Jean-Beru: As mentioned in stamp and watermark PRs, I don't this that this interface (mostly its methods) is useful. If I create my own GotenbergPdf class, I might not need to implement all these methods.

src/Builder/Behaviors/RotateTrait.php
```diff
@@ -0,0 +1,64 @@
...
+    #[WithConfigurationNode(new ScalarNodeBuilder('rotate_pages'))]
+    public function rotatePages(string|null $rotatePages = null): static
+    {
+        if (null === $rotatePages) {
+            $this->getBodyBag()->unset('rotatePages');
+
+            return $this;
+        }
+
+        $this->getBodyBag()->set('rotatePages', $rotatePages);
```
Jean-Beru: We can validate the range.

docs/pdf/HtmlPdfBuilder.md
Jean-Beru: The API seems limited to 1 rotation type for 1 range: https://gotenberg.dev/docs/manipulate-pdfs/rotate-pdfs

Neirda24 (approved):

.gitignore
```diff
@@ -8,3 +8,4 @@
 .phpunit.result.cache
 composer.lock
 phpstan.neon
+/config/reference.php
```
Jean-Beru: Should be at line 2 IMO

Jean-Beru (approved):

.gitignore
StevenRenaux: done

PR #256, opened 2026-04-10
feat: Add downloadFrom new fields (release 8.28)

StevenRenaux:
| New feature ?           | yes |

## Description

https://github.com/gotenberg/gotenberg/releases/tag/v8.28.0

> Download From: Extended the downloadFrom JSON schema with a field property ("watermark", "stamp", "embedded", > or "") to route downloaded files to the appropriate form field bucket. The existing embedded boolean is preserved for 
> backward compatibility.

### DIFF
--- a/src/Builder/Behaviors/DownloadFromTrait.php
+++ b/src/Builder/Behaviors/DownloadFromTrait.php
@@ -8,6 +8,7 @@ use Sensiolabs\GotenbergBundle\Builder\Behaviors\Dependencies\LoggerAwareTrait;
 use Sensiolabs\GotenbergBundle\Builder\BodyBag;
 use Sensiolabs\GotenbergBundle\Builder\Util\NormalizerFactory;
 use Sensiolabs\GotenbergBundle\Builder\Util\ValidatorFactory;
+use Sensiolabs\GotenbergBundle\Enumeration\DownloadFromField;
 use Sensiolabs\GotenbergBundle\NodeBuilder\ArrayNodeBuilder;
 use Sensiolabs\GotenbergBundle\NodeBuilder\EnumNodeBuilder;
 use Sensiolabs\GotenbergBundle\NodeBuilder\ScalarNodeBuilder;
@@ -21,7 +22,7 @@ trait DownloadFromTrait
     /**
      * Sets download from to download each entry (file) in parallel (URLs MUST return a Content-Disposition header with a filename parameter.).
      *
-     * @param list<array{url: string, extraHttpHeaders?: array<string, string>, field?: ''|'watermark'|'stamp'|'embedded'}> $downloadFrom
+     * @param list<array{url: string, extraHttpHeaders?: array<string, string>, field?: DownloadFromField|string}> $downloadFrom
      *
      * @see https://gotenberg.dev/docs/webhook-download#download-from
      *
@@ -33,7 +34,7 @@ trait DownloadFromTrait
             new ScalarNodeBuilder('name', required: true),
             new ScalarNodeBuilder('value', required: true),
         ]),
-        new EnumNodeBuilder('field', values: ['', 'watermark', 'stamp', 'embedded']),
+        new EnumNodeBuilder('field', callback: DownloadFromField::class),
     ]))]
     public function downloadFrom(array $downloadFrom): static
     {
@@ -67,6 +68,6 @@ trait DownloadFromTrait
     #[NormalizeGotenbergPayload]
     private function normalizeDownloadFrom(): \Generator
     {
-        yield 'downloadFrom' => NormalizerFactory::json(false);
+        yield 'downloadFrom' => NormalizerFactory::downloadFrom();
     }
 }
--- a/src/Builder/Behaviors/DownloadFromTrait.php
+++ b/src/Builder/Behaviors/DownloadFromTrait.php
@@ -9,6 +9,7 @@ use Sensiolabs\GotenbergBundle\Builder\BodyBag;
 use Sensiolabs\GotenbergBundle\Builder\Util\NormalizerFactory;
 use Sensiolabs\GotenbergBundle\Builder\Util\ValidatorFactory;
 use Sensiolabs\GotenbergBundle\NodeBuilder\ArrayNodeBuilder;
+use Sensiolabs\GotenbergBundle\NodeBuilder\EnumNodeBuilder;
 use Sensiolabs\GotenbergBundle\NodeBuilder\ScalarNodeBuilder;
 
 trait DownloadFromTrait
@@ -20,7 +21,7 @@ trait DownloadFromTrait
     /**
      * Sets download from to download each entry (file) in parallel (URLs MUST return a Content-Disposition header with a filename parameter.).
      *
-     * @param list<array{url: string, extraHttpHeaders?: array<string, string>}> $downloadFrom
+     * @param list<array{url: string, extraHttpHeaders?: array<string, string>, field?: ''|'watermark'|'stamp'|'embedded'}> $downloadFrom
      *
      * @see https://gotenberg.dev/docs/webhook-download#download-from
      *
@@ -32,6 +33,7 @@ trait DownloadFromTrait
             new ScalarNodeBuilder('name', required: true),
             new ScalarNodeBuilder('value', required: true),
         ]),
+        new EnumNodeBuilder('field', values: ['', 'watermark', 'stamp', 'embedded']),
     ]))]
     public function downloadFrom(array $downloadFrom): static
     {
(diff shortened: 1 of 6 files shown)


src/Builder/Util/ValidatorFactory.php
```diff
@@ -62,14 +62,20 @@ public static function filesExtension(array $files, array $validExtensions): voi
...
     public static function download(array $downloadFrom): void
     {
+        $validFields = ['watermark', 'stamp', 'embedded', ''];
+
         foreach ($downloadFrom as $file) {
             if (!\array_key_exists('url', $file)) {
                 throw new InvalidBuilderConfiguration('"url" is mandatory into "downloadFrom" array field.');
             }
+
+            if (\array_key_exists('field', $file) && !\in_array($file['field'], $validFields, true)) {
```
Neirda24: How does it behave if `$file['field']` is null ?

src/Builder/Behaviors/DownloadFromTrait.php
```diff
@@ -20,7 +21,7 @@ abstract protected function getBodyBag(): BodyBag;
     /**
      * Sets download from to download each entry (file) in parallel (URLs MUST return a Content-Disposition header with a filename parameter.).
      *
-     * @param list<array{url: string, extraHttpHeaders?: array<string, string>}> $downloadFrom
+     * @param list<array{url: string, extraHttpHeaders?: array<string, string>, field?: ''|'watermark'|'stamp'|'embedded'}> $downloadFrom
```
Jean-Beru: Note for next major version, we could propose constants to improve DX. (constant usage in traits was introduced in PHP 8.2 but Symfony 6.4 supports `>=8.1`).

src/Builder/Util/ValidatorFactory.php
```diff
@@ -62,14 +62,30 @@ public static function filesExtension(array $files, array $validExtensions): voi
...
                 throw new InvalidBuilderConfiguration('"url" is mandatory into "downloadFrom" array field.');
             }
+
+            if (!\is_string($file['url'])) {
+                throw new InvalidBuilderConfiguration('"url" in "downloadFrom" must be a string.');
+            }
+
+            if (\array_key_exists('field', $file)) {
+                if (!\is_string($file['field'])) {
+                    throw new InvalidBuilderConfiguration('"field" in "downloadFrom" must be a string.');
```
Jean-Beru:
```suggestion
                    throw new InvalidBuilderConfiguration(sprintf('Unsupported "downloadFrom[%d].field", expected one of "%s".', $i, implode('", "', $allowedValues)));
```

src/Builder/Util/ValidatorFactory.php
```diff
@@ -62,14 +62,30 @@ public static function filesExtension(array $files, array $validExtensions): voi
...
+                }
+
+                if (!\in_array($file['field'], $validFields, true)) {
+                    throw new InvalidBuilderConfiguration(\sprintf('Invalid "field" value "%s" in "downloadFrom". Allowed values are: "watermark", "stamp", "embedded", "".', $file['field']));
```
Jean-Beru:
```suggestion
                    throw new InvalidBuilderConfiguration(sprintf('Unsupported "downloadFrom[%d].field" "%s", expected one of "%s".', $i, $file['field'], implode('", "', $allowedValues)));
```

src/Builder/Util/ValidatorFactory.php
```diff
@@ -62,14 +62,30 @@ public static function filesExtension(array $files, array $validExtensions): voi
...
 
     /**
-     * @param list<array{url: string, extraHttpHeaders?: array<string, string>}> $downloadFrom
+     * @param list<array{url: string, extraHttpHeaders?: array<string, string>, field?: string}> $downloadFrom
      */
     public static function download(array $downloadFrom): void
     {
+        $validFields = ['watermark', 'stamp', 'embedded', ''];
+
         foreach ($downloadFrom as $file) {
```
Jean-Beru:
```suggestion
        foreach ($downloadFrom as $i => $file) {
```

src/Builder/Behaviors/DownloadFromTrait.php
Neirda24: If it is a fixed list maybe we could create an enum directly ? It would help giving sense to `''`

Neirda24 (approved):
After fixes asked form Hubert 👍

src/Builder/Behaviors/DownloadFromTrait.php
Jean-Beru: Of course! :100:

StevenRenaux: Not totally agree, if we do that we need a special normalizer. And if we do that, we need to also improve this logic on some others node as :

https://github.com/sensiolabs/GotenbergBundle/blob/1.x/src/Builder/Behaviors/Chromium/CookieTrait.php#L44

https://github.com/sensiolabs/GotenbergBundle/blob/1.x/src/Builder/Behaviors/MetadataTrait.php#L59

https://github.com/sensiolabs/GotenbergBundle/blob/1.x/src/Builder/Behaviors/WebhookTrait.php#L59

Jean-Beru: We already have a `EnumNormalizer` to convert Enum to string. 
About traits you listed, Metadata key can be anything (see https://github.com/sensiolabs/GotenbergBundle/pull/245) and Cookie/Webhook are not related to Gotenberg

StevenRenaux: Used in this context when the value is an argument to the method, and also using NativeEnumNodeBuilder in the configuration node.

In our case, to do this, during JSON normalization, we need to check if it's an enumeration, and then call the corresponding method in the supplement.

Jean-Beru: And a dedicated `NormalizerFactory::download` ?

StevenRenaux: done

src/Builder/Util/NormalizerFactory.php
```diff
@@ -49,6 +50,28 @@ public static function unit(): \Closure
         };
     }
 
+    /**
+     * @return (\Closure(string, array<string, array{url: string, extraHttpHeaders?: array<string, string>, field?: DownloadFromField|string}>): list<array<string, string>>)
```
Jean-Beru, line 54: Why also accepting a string ?

src/Builder/Util/ValidatorFactory.php
```diff
@@ -62,14 +63,26 @@ public static function filesExtension(array $files, array $validExtensions): voi
...
             if (!\array_key_exists('url', $file)) {
                 throw new InvalidBuilderConfiguration('"url" is mandatory into "downloadFrom" array field.');
             }
+
+            if (!\is_string($file['url'])) {
+                throw new InvalidBuilderConfiguration('"url" in "downloadFrom" must be a string.');
+            }
+
+            if (\array_key_exists('field', $file) && !$file['field'] instanceof DownloadFromField) {
+                if (!\in_array($file['field'], $allowedValues, true)) {
```
Jean-Beru: We can use https://www.php.net/manual/fr/backedenum.tryfrom.php here.

src/Builder/Util/NormalizerFactory.php
StevenRenaux: Because of the semantic configuration

src/Builder/Util/ValidatorFactory.php
StevenRenaux: You right 😉

src/Builder/Util/NormalizerFactory.php
Jean-Beru: `EnumNode` doesn't convert the value to an Enum before calling the method?

StevenRenaux: I don't think our configuration works with nested configuration node

StevenRenaux: Yeah into the builder we only check the parent node. And we use it into BuilderConfigurator for the arguments

Jean-Beru: Ok thanks for the clarification

Jean-Beru (approved):

StevenRenaux:
@Jean-Beru ready to merge, the error is the same as [‎src/DependencyInjection/SensiolabsGotenbergExtension.php‎](https://github.com/sensiolabs/GotenbergBundle/pull/271/changes#diff-76542cd9c89d934e7fcb7900b27518a6eacdeb0bd5b6246b7bc6380966ede471)

PR #259, opened 2026-04-17
fix: disable "declare_strict_types" php-cs-fixer rule

Jean-Beru:

## Description

Disable "declare_strict_types" php-cs-fixer rule to make CI green again. Force `strict_type` or not could be decided later.


StevenRenaux (approved):

Neirda24:
I would be more in favor of not allowing strict_types the more I think of it.

Jean-Beru:
What will be the pros/cons if we remove the strict mode? AFAIK, PHPStan already ensures this bundle does not rely on type coercion.

PR #263, opened 2026-04-29
feat: add addWebhookExtraHeaders to merge with existing webhook headers

### LINKED ISSUE #248: [Feature] Adding `webhookAddExtraHeaders` for pdf builder classes.
Is it possible to get an 'add extra headers' to the webhook request without overwriting the currently existing ones? That way you can, say, define an authorization header for the webhook in the config yaml, and then if there's info that needs to be passed to my webhook via headers, then that can be applied in a service without needing to pass in your auth string to your service.

Example config:
```
sensiolabs_gotenberg:
    http_client: 'gotenberg.client'
    webhook: 
        default:
          extra_http_headers: 
                name: 'Authorization'
                value: '%env(AuthPassword)%'
```
And in my service
```
public function myFunc(GotenbergPdfInterface $service)
{
...
$builder->webhookAddExtraHeaders(['X-MyVal-1' => 'foo', 'X-MyVal-2' => 'bar'])
...
}
```
and have the resulting header as: 
`Gotenberg-Webhook-Extra-Http-Headers: {"Authorization":"Bearer Blahblahblah","X-MyVal-1":"foo","X-MyVal-2":"bar"}`
Right now having stuff in the config and calling `$builder->webhookExtraHeaders()` overwrites any of the existing webhook extra headers.  This exists with regular extra http headers with `$builder->extraHttpHeaders()` and `$builder->addExtraHttpHeaders()` and just having that option for Webhook extra headers as well.

Thank You.

Jean-Beru:
| Issues                  | Fix #248 |

## Description

Adds `addWebhookExtraHeaders(array $extraHttpHeaders): static` to `WebhookTrait`, mirroring the existing `extraHttpHeaders()` / `addExtraHttpHeaders()` pair (overwrite vs. merge).

Use case: when webhook headers are already defined in bundle config (e.g. a static `Authorization`), callers can now append per-request headers (e.g. a correlation id) without replacing the whole set. Today, `webhookExtraHeaders()` always overwrites, so any caller-provided values wipe the configured ones.

The implementation reads the current `Gotenberg-Webhook-Extra-Http-Headers` header (stored as a JSON string in the headers bag), `array_merge`s the new headers on top, and re-encodes by delegating to `webhookExtraHeaders()`. An empty array short-circuits as a no-op, matching `addExtraHttpHeaders()`.

All 13 builders that compose `WebhookTrait` (10 PDF + 3 Screenshot) inherit the method via trait composition. Three test cases in `WebhookTestCaseTrait` (merge, empty no-op, no prior headers) are therefore exercised against every builder.

Note on naming: the issue suggests `webhookAddExtraHeaders`; this PR uses `addWebhookExtraHeaders` instead, to stay consistent with the existing `extraHttpHeaders` / `addExtraHttpHeaders` convention. Happy to rename if you'd prefer the original spelling.

### DIFF
--- a/src/Builder/Behaviors/WebhookTrait.php
+++ b/src/Builder/Behaviors/WebhookTrait.php
@@ -216,6 +216,25 @@ trait WebhookTrait
     }
 
     /**
+     * Adds extra headers to the ones already provided to the webhook endpoint, preserving previously set values.
+     *
+     * @param array<string, string> $extraHttpHeaders
+     *
+     * @example addWebhookExtraHeaders(['X-Custom-Header' => 'CustomValue'])
+     */
+    public function addWebhookExtraHeaders(array $extraHttpHeaders): static
+    {
+        if ([] === $extraHttpHeaders) {
+            return $this;
+        }
+
+        $current = $this->getHeadersBag()->get('Gotenberg-Webhook-Extra-Http-Headers');
+        $current = null !== $current ? json_decode($current, true) : [];
+
+        return $this->webhookExtraHeaders(array_merge($current, $extraHttpHeaders));
+    }
+
+    /**
      * Sets the webhook route with params and method for cases of success.
      *
      * @param string                    $route      #Route
--- a/docs/pdf/ConvertPdfBuilder.md
+++ b/docs/pdf/ConvertPdfBuilder.md
@@ -48,6 +48,7 @@ class YourController
 - [flatten](#flattenbool-bool)
 - [pdfFormat](#pdfformatsensiolabsgotenbergbundleenumerationpdfformat-format)
 - [pdfUniversalAccess](#pdfuniversalaccessbool-bool)
+- [addWebhookExtraHeaders](#addwebhookextraheadersarray-extrahttpheaders)
 - [webhook](#webhookarray-webhook)
 - [webhookConfiguration](#webhookconfigurationstring-name)
 - [webhookErrorRoute](#webhookerrorroutestring-route-array-parameters-string-method)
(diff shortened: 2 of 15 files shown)


Neirda24 (approved):

src/Builder/Behaviors/WebhookTrait.php
```diff
@@ -161,6 +161,25 @@ public function webhookExtraHeaders(array $extraHttpHeaders): static
...
+     * @example addWebhookExtraHeaders(['X-Custom-Header' => 'CustomValue'])
+     */
+    public function addWebhookExtraHeaders(array $extraHttpHeaders): static
+    {
+        if ([] === $extraHttpHeaders) {
+            return $this;
+        }
+
+        $current = $this->getHeadersBag()->get('Gotenberg-Webhook-Extra-Http-Headers');
+        $current = null !== $current ? json_decode($current, true) : [];
```
StevenRenaux, line 232: Just noticed during my PR  webhook event but we don't have in this trait a normalize mathode with #[NormalizeGotenbergPayload]. This could be fix the problem of the json_encode an decode?

Jean-Beru: Good catch ! It could be a great improvement. It could be done in a specific PR since the `#[NormalizeGotenbergPayload]` is dedicated to the request body, not headers.

StevenRenaux: you right, my bad :D

StevenRenaux (approved):

PR #265, opened 2026-04-30
feat: Allow to render Twig templates with a specific locale

Jean-Beru:
## Summary
- Add a `locale(string)` method on builders using Twig templates (`HtmlPdfBuilder`, `MarkdownPdfBuilder`, `UrlPdfBuilder` and their Screenshot counterparts) to render content, header and footer templates with the given locale via Symfony's `LocaleSwitcher`.
- Behavior is unchanged when `locale()` is not called: `symfony/translation` remains an optional dependency. A clear `LogicException` is thrown if `locale()` is called without it installed.
- Exposed as a YAML configuration node via `WithConfigurationNode` so it can be set globally per builder.

## Credits
Original idea by @StevenRenaux.

### DIFF
--- a/src/Builder/Behaviors/Chromium/ContentTrait.php
+++ b/src/Builder/Behaviors/Chromium/ContentTrait.php
@@ -5,6 +5,7 @@ namespace Sensiolabs\GotenbergBundle\Builder\Behaviors\Chromium;
 use Sensiolabs\GotenbergBundle\Builder\Attributes\NormalizeGotenbergPayload;
 use Sensiolabs\GotenbergBundle\Builder\Attributes\WithConfigurationNode;
 use Sensiolabs\GotenbergBundle\Builder\Behaviors\Dependencies\AssetBaseDirFormatterAwareTrait;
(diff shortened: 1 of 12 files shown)


StevenRenaux (approved):

src/Builder/Behaviors/Chromium/ContentTrait.php
```diff
@@ -187,6 +207,23 @@ public function footerFile(string $path): static
      * @throws PartRenderingException if the template could not be rendered
      */
     protected function withRenderedPart(Part $part, string $template, array $context = []): static
+    {
+        if (null !== $this->locale && $this->getLocaleSwitcher()->getLocale() !== $this->locale) {
+            return $this->getLocaleSwitcher()->runWithLocale(
```
Neirda24, line 212: what about using locale aware instead ?

Jean-Beru: Using `LocaleAwareTrait` to retrieve the current locale? I’m not sure that meets the goal of this PR: enabling PDF generation in multiple languages.

Neirda24 (changes requested):
Depends on https://github.com/sensiolabs/GotenbergBundle/pull/279

PR #266, opened 2026-05-06
fix: remove content* methods from UrlPdfBuilder and UrlScreenshotBuilder

### LINKED ISSUE #257: [Bug] Remove content* methods from UrlBuilder
According to the documentation, both URL, HTML and Markdown needs HTML for their header and footer. On the other hand, `content*` methods should not be present in the URL builder (Markdown still need it [as a wrapper](https://gotenberg.dev/docs/convert-with-chromium/convert-markdown-to-pdf#basics) for md files).

_Originally posted by @Jean-Beru in https://github.com/sensiolabs/GotenbergBundle/pull/251#discussion_r3071845937_

marilenaRM:
| Issues                  | Fix #257 |

## Description

For URL-based builders, the page body is provided by the URL itself. Exposing content(), contentRaw() and contentFile() was misleading since those methods have no effect on URL conversions.

Superseded by #274

### DIFF
--- a/UPGRADE-1.2.md
+++ b/UPGRADE-1.2.md
@@ -1,1 +1,13 @@
 # UPGRADE FROM 1.1.0 to 1.2.0
+
+## Deprecations
+
+### `ContentTrait`: header and footer methods moved to `PageMarginalTrait`
+
+The `header()`, `headerRaw()`, `headerFile()`, `footer()`, `footerRaw()` and `footerFile()` methods have been moved out of `ContentTrait` into the new dedicated `PageMarginalTrait`. These methods are still accessible through `ContentTrait` in 1.2 but will be removed in 2.0.
+
+If you use `ContentTrait` directly in your own builders, add `PageMarginalTrait` instead to handle headers and footers.
+
+### `UrlPdfBuilder` and `UrlScreenshotBuilder`: `content*` methods deprecated
+
+The `content()`, `contentRaw()` and `contentFile()` methods are deprecated on `UrlPdfBuilder` and `UrlScreenshotBuilder`. Since the page body is provided by the URL itself, these methods have no effect. Use `url()` or `route()` instead. They will be removed in 2.0.
--- a/docs/generate.php
+++ b/docs/generate.php
@@ -131,6 +131,7 @@ class BuilderParser
 
         $this->prepareBuilderFromClass($class);
         $this->cleanBuilderFromClass($class);
+        $this->filterDeprecatedMethods($class);
 
         foreach ($this->parts['methods']['@'] as $methodName => $parts) {
             $package = $parts['package'] ?? '@';
@@ -341,7 +342,7 @@ class BuilderParser
                 $this->parts['methods']['@'][$method->getShortName()]['package'] = $newPackage;
             }
 
-            if (isset($parsedDocBlock['description']) && [''] !== $parsedDocBlock['description']) {
+            if (isset($parsedDocBlock['description']) && [] !== $parsedDocBlock['description'] && [''] !== $parsedDocBlock['description']) {
                 $this->parts['methods']['@'][$method->getShortName()]['description'] = $parsedDocBlock['description'];
             }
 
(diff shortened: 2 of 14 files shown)


Jean-Beru (commented):
Thanks you @marilenaRM !

It could be considered as a BC break but it's a bug fix... Maybe deprecating these methods will be more appropriated. WDYT @Neirda24 @StevenRenaux 

PS: don't forget to update `UPGRADE`  file to mention this change.

StevenRenaux:
> Thanks you @marilenaRM !
> 
> It could be considered as a BC break but it's a bug fix... Maybe deprecating these methods will be more appropriated. WDYT @Neirda24 @StevenRenaux 
> 
> PS: don't forget to update `UPGRADE`  file to mention this change.

It would work I would say yes to the BC break but as it is not the case, not sure of the usefulness

Neirda24:
Shouldn't we split this in a dedicated trait ? As experienced before in this project we noticed that handling trait conflict while using reflection might just get things messy. Not in this specific case though.

Don't mind the BC it is a bugfix to me as it doesn't produce anything I guess.

marilenaRM:
Hello ! :)

@Jean-Beru : I just added a commit for the update of the file UPGRADE :)
@Neirda24 and @StevenRenaux the BC break notification seems better since we remove some (most certainly not used) public functions, so in my opinion it should be noticed.

 @Neirda24 I pushed a separate commit for a way to introduce a separated Trait, WDYT?

Jean-Beru:
> Don't mind the BC it is a bugfix to me as it doesn't produce anything I guess.

That's the reason why it can break current code. The following code works even if the `content` method is ignored. Removing it will lead to fatal a error.
```php
$builder
    ->content(/*...*/)
    ->url(/* ... */)
```

> Shouldn't we split this in a dedicated trait ?

:+1: 

`ContentTrait` can be split in 2 traits: content vs. header/footer (a.k.a. marginal ?).

Then we could:
- deprecated `content*` methods in `UrlPdfBuilder`
- deprecate `header*` and `footer*` methods in `ContentTrait` and invite to use `MarginalTrait` instead
- use `MarginalTrait` in our builder to override the deprecated `header*` and `footer*` methods

Later we could:
- remove `ContentTrait` usage in `UrlPdfBuilder` and `MarkdownPdfBuilder`
- remove  `header*` and `footer*` methods in `ContentTrait`

StevenRenaux:
> > Don't mind the BC it is a bugfix to me as it doesn't produce anything I guess.
> 
> That's the reason why it can break current code. The following code works even if the `content` method is ignored. Removing it will lead to fatal a error.
> 
> ```
> $builder
>     ->content(/*...*/)
>     ->url(/* ... */)
> ```
> 
> > Shouldn't we split this in a dedicated trait ?
> 
> 👍
> 
> `ContentTrait` can be split in 2 traits: content vs. header/footer (a.k.a. marginal ?).
> 
> Then we could:
> 
> * deprecated `content*` methods in `UrlPdfBuilder`
> * deprecate `header*` and `footer*` methods in `ContentTrait` and invite to use `MarginalTrait` instead
> * use `MarginalTrait` in our builder to override the deprecated `header*` and `footer*` methods
> 
> Later we could:
> 
> * remove `ContentTrait` usage in `UrlPdfBuilder` and `MarkdownPdfBuilder`
> * remove  `header*` and `footer*` methods in `ContentTrait`

This solution LGTM :star_struck:

marilenaRM:
so, what's the plan? deprecate in 1.2 and BC Break in 2.0? 
maybe we should also address the usage of header and footer in ScreenShots in the referenced issue, deprecate them as well? I'll roll back my changes and split the Traits responsabilities, if that's ok with you

Jean-Beru:
> so, what's the plan? deprecate in 1.2 and BC Break in 2.0? maybe we should also address the usage of header and footer in ScreenShots in the referenced issue, deprecate them as well? I'll roll back my changes and split the Traits responsabilities, if that's ok with you

LGTM

PR #274, opened 2026-05-19
Refactor/deprecate content methods

### LINKED ISSUE #257: [Bug] Remove content* methods from UrlBuilder

marilenaRM:
| Bug fix ?               | yes  + Refactor|
| BC break ?              | no  |
| Issues                  | Fix #257 |

## Description

  Summary
  
  - Extract header/footer methods from ContentTrait into a new dedicated PageMarginalTrait; ChromiumPdfTrait resolves the conflict via insteadof so PDF builders transparently use the real implementation
  - Deprecate content(), contentRaw() and contentFile() on UrlPdfBuilder and UrlScreenshotBuilder via a new DeprecatedBodyContentTrait (these builders receive their body from the URL, not from injected HTML)
  - Fix three bugs in docs/generate.php: @example tags were silently dropped when a method was seen via multiple traits, an empty description from a @deprecated-only docblock could overwrite a non-empty one, and deprecated
   no-op stubs appeared in the generated docs without any content

  See UPGRADE-1.2.md for migration instructions.

Replaces #266 

  Test plan

  - ./vendor/bin/phpunit — all tests pass
  - ./vendor/bin/phpstan analyse — no new errors
  - ./vendor/bin/php-cs-fixer fix --diff — no changes
  - dagger call generate-docs export --path . — docs regenerated cleanly, deprecated methods absent from output
  - Confirm HtmlPdfBuilder::header() uses PageMarginalTrait (not the deprecated stub)
  - Confirm UrlPdfBuilder::content() fires E_USER_DEPRECATED

### DIFF
--- a/UPGRADE-1.2.md
+++ b/UPGRADE-1.2.md
@@ -1,1 +1,13 @@
 # UPGRADE FROM 1.1.0 to 1.2.0
+
+## Deprecations
+
+### `ContentTrait`: header and footer methods moved to `PageMarginalTrait`
+
+The `header()`, `headerRaw()`, `headerFile()`, `footer()`, `footerRaw()` and `footerFile()` methods have been moved out of `ContentTrait` into the new dedicated `PageMarginalTrait`. These methods are still accessible through `ContentTrait` in 1.2 but will be removed in 2.0.
+
+If you use `ContentTrait` directly in your own builders, add `PageMarginalTrait` instead to handle headers and footers.
+
+### `UrlPdfBuilder` and `UrlScreenshotBuilder`: `content*` methods deprecated
+
+The `content()`, `contentRaw()` and `contentFile()` methods are deprecated on `UrlPdfBuilder` and `UrlScreenshotBuilder`. Since the page body is provided by the URL itself, these methods have no effect. Use `url()` or `route()` instead. They will be removed in 2.0.
--- a/UPGRADE-1.2.md
+++ b/UPGRADE-1.2.md
@@ -1,13 +1,1 @@
 # UPGRADE FROM 1.1.0 to 1.2.0
-
-## Deprecations
-
-### `ContentTrait`: header and footer methods moved to `PageMarginalTrait`
-
-The `header()`, `headerRaw()`, `headerFile()`, `footer()`, `footerRaw()` and `footerFile()` methods have been moved out of `ContentTrait` into the new dedicated `PageMarginalTrait`. These methods are still accessible through `ContentTrait` in 1.2 but will be removed in 2.0.
-
-If you use `ContentTrait` directly in your own builders, add `PageMarginalTrait` instead to handle headers and footers.
-
-### `UrlPdfBuilder` and `UrlScreenshotBuilder`: `content*` methods deprecated
-
-The `content()`, `contentRaw()` and `contentFile()` methods are deprecated on `UrlPdfBuilder` and `UrlScreenshotBuilder`. Since the page body is provided by the URL itself, these methods have no effect. Use `url()` or `route()` instead. They will be removed in 2.0.
--- a/docs/screenshot/HtmlScreenshotBuilder.md
+++ b/docs/screenshot/HtmlScreenshotBuilder.md
@@ -107,12 +107,6 @@ class YourController
 - [content](#contentstring-template-array-context)
 - [contentFile](#contentfilestring-path)
 - [contentRaw](#contentrawstring-html)
-- [footer](#footerstring-template-array-context)
-- [footerFile](#footerfilestring-path)
-- [footerRaw](#footerrawstring-html)
-- [header](#headerstring-template-array-context)
-- [headerFile](#headerfilestring-path)
-- [headerRaw](#headerrawstring-html)
 - [emulatedMediaFeatures](#emulatedmediafeaturesarray-emulatedmediafeatures)
 - [failOnConsoleExceptions](#failonconsoleexceptionsbool-bool)
 - [failOnHttpStatusCodes](#failonhttpstatuscodesarray-statuscodes)
@@ -492,30 +486,6 @@ return $gotenberg
     ->stream()
 ;
 ```
-
-### footer(string \$template, array \$context)
-> [!TIP]
-> See: [https://gotenberg.dev/docs/convert-with-chromium/convert-html-to-pdf#header--footer](https://gotenberg.dev/docs/convert-with-chromium/convert-html-to-pdf#header--footer)
-
-### footerFile(string \$path)
-> [!TIP]
-> See: [https://gotenberg.dev/docs/convert-with-chromium/convert-html-to-pdf#header--footer](https://gotenberg.dev/docs/convert-with-chromium/convert-html-to-pdf#header--footer)
-
-### footerRaw(string \$html)
-> [!TIP]
-> See: [https://gotenberg.dev/docs/convert-with-chromium/convert-html-to-pdf#header--footer](https://gotenberg.dev/docs/convert-with-chromium/convert-html-to-pdf#header--footer)
-
-### header(string \$template, array \$context)
-> [!TIP]
-> See: [https://gotenberg.dev/docs/convert-with-chromium/convert-html-to-pdf#header--footer](https://gotenberg.dev/docs/convert-with-chromium/convert-html-to-pdf#header--footer)
-
-### headerFile(string \$path)
-> [!TIP]
-> See: [https://gotenberg.dev/docs/convert-with-chromium/convert-html-to-pdf#header--footer](https://gotenberg.dev/docs/convert-with-chromium/convert-html-to-pdf#header--footer)
-
-### headerRaw(string \$html)
-> [!TIP]
-> See: [https://gotenberg.dev/docs/convert-with-chromium/convert-html-to-pdf#header--footer](https://gotenberg.dev/docs/convert-with-chromium/convert-html-to-pdf#header--footer)
 
 
 ### emulatedMediaFeatures(array \$emulatedMediaFeatures)
--- a/docs/screenshot/HtmlScreenshotBuilder.md
+++ b/docs/screenshot/HtmlScreenshotBuilder.md
@@ -107,6 +107,12 @@ class YourController
 - [content](#contentstring-template-array-context)
 - [contentFile](#contentfilestring-path)
 - [contentRaw](#contentrawstring-html)
+- [footer](#footerstring-template-array-context)
+- [footerFile](#footerfilestring-path)
+- [footerRaw](#footerrawstring-html)
+- [header](#headerstring-template-array-context)
+- [headerFile](#headerfilestring-path)
+- [headerRaw](#headerrawstring-html)
 - [emulatedMediaFeatures](#emulatedmediafeaturesarray-emulatedmediafeatures)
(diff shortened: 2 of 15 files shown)


src/Builder/Behaviors/Chromium/ContentTrait.php
```diff
@@ -71,113 +71,93 @@ public function contentFile(string $path): self
...
-     * @example header('header.html.twig', ['my_var' => 'value'])
      */
     #[WithConfigurationNode(new ArrayNodeBuilder('header', children: [
         new ScalarNodeBuilder('template', required: true, restrictTo: 'string'),
         new ArrayNodeBuilder('context', normalizeKeys: false, prototype: 'variable'),
     ]))]
     public function header(string $template, array $context = []): static
     {
-        return $this->withRenderedPart(Part::Header, $template, $context);
+        @trigger_error(\sprintf('Since sensiolabs/gotenberg-bundle 1.2: "%s" is deprecated and will be removed in 2.0. Header/footer support has been moved to "%s"; this builder does not support it.', __METHOD__, PageMarginalTrait::class), \E_USER_DEPRECATED);
```
Jean-Beru: What about https://www.php.net/manual/en/class.deprecated.php ?

Ex:
```php
#[Deprecated(message:'Will be removed in 2.0. Use Sensiolabs\GotenbergBundle\Builder\Behaviors\Chromium\PageMarginalTrait instead', since:'sensiolabs/GotenbergBundle:1.3')]

```

docs/screenshot/HtmlScreenshotBuilder.md
```diff
@@ -107,12 +107,6 @@ class YourController
 - [content](#contentstring-template-array-context)
 - [contentFile](#contentfilestring-path)
 - [contentRaw](#contentrawstring-html)
-- [footer](#footerstring-template-array-context)
```
Neirda24, line 110: weird... why would header / footer doesn't work ?

src/Builder/Behaviors/Chromium/ContentTrait.php
marilenaRM: done!

src/Builder/Behaviors/Chromium/DeprecatedBodyContentTrait.php
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Sensiolabs\GotenbergBundle\Builder\Behaviors\Chromium;
+
+/**
+ * @package Behavior\\Content
+ */
+trait DeprecatedBodyContentTrait
+{
+    /**
+     * @deprecated since 1.2, will be removed in 2.0. The page body is provided by the URL.
```
Jean-Beru:
```suggestion
     * @deprecated since 1.4, will be removed in 2.0. The page body is provided by the URL.
```

This also applies to the following annotations/triggers.

UPGRADE-1.2.md
```diff
@@ -1 +1,13 @@
...
+
+### `ContentTrait`: header and footer methods moved to `PageMarginalTrait`
+
+The `header()`, `headerRaw()`, `headerFile()`, `footer()`, `footerRaw()` and `footerFile()` methods have been moved out of `ContentTrait` into the new dedicated `PageMarginalTrait`. These methods are still accessible through `ContentTrait` in 1.2 but will be removed in 2.0.
+
+If you use `ContentTrait` directly in your own builders, add `PageMarginalTrait` instead to handle headers and footers.
+
+### `UrlPdfBuilder` and `UrlScreenshotBuilder`: `content*` methods deprecated
+
+The `content()`, `contentRaw()` and `contentFile()` methods are deprecated on `UrlPdfBuilder` and `UrlScreenshotBuilder`. Since the page body is provided by the URL itself, these methods have no effect. Use `url()` or `route()` instead. They will be removed in 2.0.
```
Jean-Beru: It should be in the `UPGRADE-1.4.md` file.

UPGRADE-1.2.md
```diff
@@ -1 +1,13 @@
 # UPGRADE FROM 1.1.0 to 1.2.0
+
+## Deprecations
+
+### `ContentTrait`: header and footer methods moved to `PageMarginalTrait`
+
+The `header()`, `headerRaw()`, `headerFile()`, `footer()`, `footerRaw()` and `footerFile()` methods have been moved out of `ContentTrait` into the new dedicated `PageMarginalTrait`. These methods are still accessible through `ContentTrait` in 1.2 but will be removed in 2.0.
```
Jean-Beru:
```suggestion
The `header()`, `headerRaw()`, `headerFile()`, `footer()`, `footerRaw()` and `footerFile()` methods have been moved out of `ContentTrait` into the new dedicated `PageMarginalTrait`. These methods are still accessible through `ContentTrait` in 1.4 but will be removed in 2.0.
```

src/Builder/Behaviors/Chromium/DeprecatedBodyContentTrait.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace Sensiolabs\GotenbergBundle\Builder\Behaviors\Chromium;
+
+/**
+ * @package Behavior\\Content
+ */
+trait DeprecatedBodyContentTrait
```
Jean-Beru, line 10:
```suggestion
#[Internal]
trait DeprecatedBodyContentTrait
```
This trait only exists for BC, we should mark it as internal to limit its impact after its removal.

src/Builder/Behaviors/Chromium/ContentTrait.php
```diff
@@ -77,104 +77,70 @@ public function contentFile(string $path): self
...
      *
      * @see https://gotenberg.dev/docs/convert-with-chromium/convert-html-to-pdf#header--footer
-     *
-     * @example header('header.html.twig', ['my_var' => 'value'])
      */
     #[WithConfigurationNode(new ArrayNodeBuilder('header', children: [
         new ScalarNodeBuilder('template', required: true, restrictTo: 'string'),
         new ArrayNodeBuilder('context', normalizeKeys: false, prototype: 'variable'),
     ]))]
+    #[\Deprecated(message: 'Will be removed in 2.0. Use Sensiolabs\GotenbergBundle\Builder\Behaviors\Chromium\PageMarginalTrait instead', since: 'sensiolabs/GotenbergBundle:1.4')]
```
Neirda24: should this trait use the PageMarginal one then ? To properly forward or would this create some bugs ?

src/Builder/Behaviors/Chromium/DeprecatedBodyContentTrait.php
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * @package Behavior\\Content
+ *
+ * @internal
+ */
+trait DeprecatedBodyContentTrait
+{
+    /**
+     * @deprecated since 1.4, will be removed in 2.0. The page body is provided by the URL.
```
Neirda24, line 13: Use the Dprecated attribute instead ?

src/Builder/Behaviors/Chromium/ContentTrait.php
marilenaRM: yeah it would :) both ContentTrait and PageMarginalTrait declare header*/footer*, so use PageMarginalTrait here is a trait-method collision (I'd need insteadof/as). It would also pull normalizePageMarginal() (the header.html/footer.html payload) and the #[WithConfigurationNode] into every ContentTrait consumer, including the screenshot builders, which must not emit header/footer. So I resolve it at the builder level instead (ChromiumPdfTrait with insteadof): the PDF builders get the real PageMarginalTrait methods, and ContentTrait just keeps the deprecated shims for BC. Forwarding would bring back exactly the trait-conflict mess from [#266](https://github.com/sensiolabs/GotenbergBundle/pull/266).

docs/screenshot/HtmlScreenshotBuilder.md
Jean-Beru: They are not in the documentation :-/
https://gotenberg.dev/docs/convert-with-chromium/screenshot-html

marilenaRM: that's actually the point :) header/footer are a PDF-only thing (page margins), and Gotenberg's screenshot endpoints don't support them. On the screenshot builders they weren't even serialized, so they had no effect. They live in PageMarginalTrait now (only the PDF builders use it), and the screenshot builders just keep the deprecated ContentTrait shims until 2.0. Since those are @deprecated, the doc generator drops them, so they're gone from the screenshot docs too.

src/Builder/Behaviors/Chromium/DeprecatedBodyContentTrait.php
marilenaRM: actually I went the other way and dropped the attribute instead. #[\Deprecated] only fires at runtime on PHP 8.4+, and we still support 8.1, where it does nothing (I checked on 8.3: the \Deprecated class isn't even there, so nothing is raised when you call the method). So both traits use @deprecated phpdoc + @trigger_error(E_USER_DEPRECATED) now. That way it actually shows up on every supported version, IDEs/PHPStan still catch it via @deprecated, and the two traits end up consistent like you wanted.

marilenaRM:
@Jean-Beru heads-up: I ended up dropping the #[\Deprecated] attribute you suggested. Turns out it only fires at runtime on PHP 8.4+, so on 8.1–8.3 (which we still support) the header/footer deprecations were silent. I checked on 8.3: the \Deprecated class doesn't even exist, so nothing is raised when the method is called. So I moved ContentTrait onto @deprecated phpdoc + @trigger_error (and kept DeprecatedBodyContentTrait the same way), which fires everywhere and still gives IDE/PHPStan detection. If you'd rather keep the attribute, we could do #[\Deprecated] + a trigger_error guarded by PHP_VERSION_ID < 80400, just more verbose across the methods. I'm fine either way, wdyt?

marilenaRM:
Closing in favour of #290.

After the review it was clear the extraction of `PageMarginalTrait` was not the right shape: it left `ContentTrait` full of no-op stubs and needed `insteadof` in `ChromiumPdfTrait` to undo them. Worse, it turned a part Gotenberg silently ignores into a hard failure — `ContentTrait::normalizeContent()` stopped yielding `header.html`, so on the three screenshot builders the raw `RenderedPart` fell back to `NormalizerFactory::noop()` and `FormDataPart` rejected it at `generate()`.

#290 follows the suggestion instead: duplicate `ContentTrait` into `PdfContentTrait` and `ScreenshotContentTrait`, mirroring the existing `PdfPagePropertiesTrait` / `ScreenshotPagePropertiesTrait` split, deprecate `ContentTrait` as a whole and deprecate the methods each route ignores. No `insteadof`, no behaviour change, and 2.0 becomes a pure deletion. `symfony/deprecation-contracts` is in as well.

Opening a new PR rather than force-pushing this one, since none of the code discussed here survives.

The docs generator fixes that were bundled here are split out into #289.

PR #276, opened 2026-05-20
feat(ci): add Gotenberg integration tests workflow and a version matrix

### LINKED ISSUE #158: [CI] Make sure new Gotenberg versions doesn't break the bundle
The goal would be to run a workflow on :
* new release of https://github.com/gotenberg/gotenberg
* new release of https://github.com/symfony/symfony
* new release of this bundle
* periodically ?

We need to create new tests for such use cases that test endpoints with all of possible properties and their variant (for width for example we can pass with or without units)

Nitram1123:
| Issues                  | Fix #158 |

## Description
This PR prepares the integration testing groundwork for issue #158.
It introduces the initial GitHub Actions integration workflow and a matrix to run integration tests against multiple Gotenberg Docker image versions in CI. It also adds a separate nightly latest compatibility check to detect upstream Docker image regressions early. So follow-up PRs can focus on expanding integration coverage.
This is a preparatory step toward adding more integration tests across the bundle’s supported builders and options.
Linked to #158.


Neirda24:
impressive work. thank you. will take a look soon.

tests/Integration/Attributes/RequiresGotenberg.php
```diff
@@ -0,0 +1,17 @@
...
+{
+    public function __construct(
+        private readonly string $versionRequirement,
+    ) {
+    }
+
+    public function versionRequirement(): string
+    {
+        return $this->versionRequirement;
+    }
```
Jean-Beru: `$versionRequirement` could be public to remove the getter.

tests/Integration/AbstractGotenbergIntegrationTestCase.php
```diff
@@ -0,0 +1,147 @@
...
+
+    protected static function requireVersionAtLeast(string $version): void
+    {
+        self::requireGotenbergVersionRequirement(\sprintf('>= %s', $version));
+    }
+
+    protected static function requireVersionBelow(string $version): void
+    {
+        self::requireGotenbergVersionRequirement(\sprintf('< %s', $version));
+    }
```
Jean-Beru: Are these methods used?

Nitram1123: Good catch. I introduced the attribute and forgot to remove these methods.

.dagger/src/IntegrationTestsGotenbergBundle.php
```diff
@@ -0,0 +1,156 @@
...
+
+        return $this->symfonyContainer
+            ->withExec(['php', '-r', $script])
+            ->stdout()
+        ;
+    }
+
+    #[DaggerFunction]
+    #[Doc('Reads the raw /version response directly from the bound Gotenberg service.')]
+    public function rawVersion(): string
```
Jean-Beru: Is this method useful?

tests/Integration/AbstractGotenbergIntegrationTestCase.php
```diff
@@ -0,0 +1,124 @@
...
+namespace Sensiolabs\GotenbergBundle\Tests\Integration;
+
+use PHPUnit\Framework\Assert;
+use Sensiolabs\GotenbergBundle\Builder\Result\GotenbergFileResult;
+use Sensiolabs\GotenbergBundle\Processor\InMemoryProcessor;
+use Sensiolabs\GotenbergBundle\Tests\Integration\Attributes\RequiresGotenberg;
+use Sensiolabs\GotenbergBundle\Version\VersionFetcherInterface;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+abstract class AbstractGotenbergIntegrationTestCase extends KernelTestCase
```
Jean-Beru: We can follow Symfony convention.
```suggestion
abstract class GotenbergIntegrationTestCase extends KernelTestCase
```

tests/Integration/GotenbergIntegrationTestCase.php
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    protected static function resolvedGotenbergVersionUnderTest(): string
+    {
+        if (null !== self::$resolvedGotenbergVersionUnderTest) {
+            return self::$resolvedGotenbergVersionUnderTest;
+        }
+
+        /** @var VersionFetcherInterface $versionFetcher */
+        $versionFetcher = static::getContainer()->get(VersionFetcherInterface::class);
```
Jean-Beru, line 43: Do you know if the Gotenberg API is called to retrieve its version? It could have a big impact on each test.

Nitram1123: The test case caches the result in a `static` property (`$resolvedGotenbergVersionUnderTest`), so the call only happens **once per PHPUnit run**, not per test.
The service itself also memoizes the response (`$this->version ??= …`).
And it's only triggered for tests carrying a `#[RequiresGotenberg]` attribute, so tests without version constraints never hit the network at all.

.dagger/src/GotenbergBundle.php
```diff
@@ -25,21 +25,48 @@ class GotenbergBundle
...
+            yield $row['gotenberg-version'];
+        }
+    }
+
+    private function gotenbergImage(
+        string $gotenbergVersion = self::DEFAULT_GOTENBERG_VERSION,
+        string $gotenbergVariant = 'full',
+    ): string {
+        return match ($gotenbergVariant) {
+            'full' => "gotenberg/gotenberg:{$gotenbergVersion}",
```
Neirda24: maybe full could be the default instead of an error ? thus allowing `null` instead ?

.dagger/src/GotenbergBundle.php
```diff
@@ -104,6 +130,27 @@ private function symfonyContainer(
         ;
     }
 
+    private function integrationSymfonyContainer(
```
Neirda24, line 133: maybe we can simplify this method to accept a "symfonycontainer" and return it after modification instead ?

.dagger/src/GotenbergBundle.php
```diff
@@ -210,4 +295,36 @@ public function testsMatrix(
...
+            $tests[] = async(fn () => $this->integrationTest(
+                $source,
+                $phpVersion,
+                $symfonyVersion,
+                $minimumStability,
+                $gotenbergVersion,
+            ));
+        }
+
+        return array_values(await($tests));
```
Neirda24: why await ? return the array directly and it will allow test to be ran concurrently

.dagger/src/integration-matrix-versions.json
```diff
@@ -0,0 +1,12 @@
+[
+    {
+        "name": "Gotenberg '8.15'",
```
Neirda24, line 3: is this the minimum version we support ?

.dagger/src/integration-matrix-versions.json
```diff
@@ -0,0 +1,12 @@
...
+        "gotenberg-version": "8.15",
+        "allow-failure": false
+    },
+    {
+        "name": "Gotenberg '8.27'",
+        "gotenberg-version": "8.27",
```
Neirda24, line 9: would be nice to add a third pointing towards "latest"

.dagger/src/IntegrationTestsGotenbergBundle.php
```diff
@@ -0,0 +1,147 @@
...
+    #[DaggerFunction]
+    #[Doc('Get the integration container for tests.')]
+    public function terminal(): Container
+    {
+        return $this->symfonyContainer->terminal();
+    }
+
+    #[DaggerFunction]
+    #[Doc('Checks that the bound Gotenberg service is reachable from the test container.')]
+    public function health(): string
```
Neirda24, line 60: is it still needed ? I guess this was for development purposes only

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,64 @@
+name: 'Integration Tests'
+
+on:
+  push:
+    branches: ['main', '1.x']
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
```
Neirda24, line 8: empty ?

tests/Integration/GotenbergIntegrationTestCase.php
```diff
@@ -0,0 +1,124 @@
...
+    {
+        parent::setUp();
+
+        self::requireIntegrationEnabled();
+        $this->requireGotenbergVersionConstraints();
+    }
+
+    protected static function requireIntegrationEnabled(): void
+    {
+        if ('1' !== getenv('GOTENBERG_INTEGRATION_ENABLED')) {
```
Neirda24, line 31: I wonder. Would a phpunit extension be best suited for this ?

.dagger/src/TestsGotenbergBundle.php
```diff
@@ -93,7 +93,7 @@ public function validateDependencies(): string
     #[Doc('Run phpunit tests and returns the container it ran in.')]
     public function phpunit(string $filter = ''): string
     {
-        $exec = ['./vendor/bin/phpunit', '--display-all-issues'];
+        $exec = ['./vendor/bin/phpunit', '--testsuite', 'default', '--display-all-issues'];
```
Neirda24, line 96: probably need to update something in `.md` files as well

Neirda24 (commented):
great stuff @Nitram1123 . Thank you. Some adjustements / questions

.dagger/src/GotenbergBundle.php
Nitram1123: I have the feeling it'd be better to keep this throwable. This value ends up in the Docker image tag, and silently falling back on a typo would hide potential bugs (fail-fast).
But, I switched $gotenbergVariant to ?string with null as default, full isn't actually a Docker Hub tag suffix — it's just the plain unsuffixed tag.

.dagger/src/integration-matrix-versions.json
Nitram1123: Nope, min supported is still 8.0. Picked 8.15 because it's right below the flatten 8.16 threshold, so both branches of VersionTest actually run instead of skipping. This PR is just to introduce the feature, follow-up PRs will add more tests and versions to the matrix.

.dagger/src/integration-matrix-versions.json
Nitram1123: Already got that, just as a separate latest job further down in the workflow instead of a matrix entry — runs nightly + on manual dispatch only, so a floating tag breaking upstream doesn't make every push non-deterministic.

.dagger/src/IntegrationTestsGotenbergBundle.php
Nitram1123: Yeah, it's dev-only, it's a minimal, fast check no Symfony kernel boot, no PHPUnit, useful when debugging. It's the same pattern TestsGotenbergBundle::all().

.github/workflows/integration-tests.yml
Nitram1123: Yeah, intentional, empty workflow_dispatch: is valid YAML, just enables manual triggering with no custom inputs. Needed so the latest job (gated on schedule/dispatch) can also be run on demand.

tests/Integration/GotenbergIntegrationTestCase.php
Nitram1123: Same mechanism PHPUnit's own docs use for conditional skipping (setUp() + markTestSkipped(), https://docs.phpunit.de/en/10.5/writing-tests-for-phpunit.html. 
Not sure that an Extension can do this anyway, the API is observer-only, no way to intercept or skip a test from it.

.dagger/src/IntegrationTestsGotenbergBundle.php
Neirda24: hmm no it is not the same pattern. `all` is to run all tests. health is just printing stuff I guess ? If it is not used I would prefer to remove it to maintain less stuff.

.github/workflows/integration-tests.yml
Neirda24: is it the same as `workflow_dispatch: ~` ? if so I would prefer the explicit null value

PR #277, opened 2026-05-29
feat: add files order control for merge endpoint

### LINKED ISSUE #264: feat: allow sorting sent files
As mentionned in https://github.com/gotenberg/gotenberg/issues/1498 by @jordy-manner, it's not possible to sort sent files.

We could add a method to change the filename in the form data (not the file itself, ex: `--form 'files=@portrait.pdf;filename=01-portrait.pdf'`) in order to respect calls order. Example:
```php
$builder
    ->files('c.pdf', 'a.pdf', 'b.pdf', 'a.pdf')
    ->sortFilesByName() // Default behavior => a, a, b, c
    ->sortFilesByCall() // Files are renamed => c, a, b, a
```

It will concern the Merge endpoint, maybe the Embed once (not sure).

Nitram1123:
Introduce sortFilesByName() and sortFilesByCall() on MergePdfBuilder to control how Gotenberg orders files in the merged PDF.

The merge endpoint sorts files alphanumerically by their multipart filename. sortFilesByCall() prefixes each multipart filename with a zero-padded counter (e.g. 000001-document.pdf) so the merge follows the order in which files() was called. sortFilesByName() restores the default behavior. Files on disk are never renamed.

Closes #264

| Issues                  | Fix #264 |

### DIFF
--- a/src/Builder/Behaviors/FilesOrderTrait.php
+++ b/src/Builder/Behaviors/FilesOrderTrait.php
@@ -1,0 +1,69 @@
+<?php
+
+namespace Sensiolabs\GotenbergBundle\Builder\Behaviors;
+
+use Psr\Log\LoggerInterface;
+use Sensiolabs\GotenbergBundle\Builder\Attributes\NormalizeGotenbergPayload;
+use Sensiolabs\GotenbergBundle\Builder\Util\NormalizerFactory;
+use Sensiolabs\GotenbergBundle\Version\Version;
+use Symfony\Component\Mime\Part\DataPart;
+use Symfony\Component\Mime\Part\File;
+
+trait FilesOrderTrait
+{
+    use FilesTrait {
+        normalizeFiles as private normalizeFilesOriginal;
+    }
+
+    /**
+     * @var 'name'|'call'
+     */
+    private string $filesSortMode = 'name';
+
+    /**
+     * Lets Gotenberg sort the files alphanumerically by their multipart filename.
+     * This is the default behavior.
+     */
+    public function sortFilesByName(): self
+    {
+        $this->filesSortMode = 'name';
+
+        return $this;
+    }
+
+    /**
+     * Preserves the order in which files were added to the builder.
+     * Each file's multipart filename is prefixed with a zero-padded counter
+     * (e.g. `000001-document.pdf`) so that Gotenberg's alphanumeric sort
+     * yields the original call order. The file on disk is not renamed.
+     */
+    public function sortFilesByCall(): self
+    {
+        $this->filesSortMode = 'call';
+
+        return $this;
+    }
+
+    #[NormalizeGotenbergPayload]
+    private function normalizeFiles(): \Generator
+    {
+        $mode = $this->filesSortMode;
+        $delegate = NormalizerFactory::asset();
+
+        yield 'files' => static function (string $key, array $assets, Version $version, LoggerInterface|null $logger) use ($mode, $delegate): \Generator {
+            if ('call' !== $mode || [] === $assets) {
+                yield from $delegate($key, $assets);
+
+                return;
+            }
+
+            $index = 1;
+            foreach ($assets as $entryKey => $info) {
+                yield ['files' => new DataPart(
+                    new File($info),
+                    \sprintf('%06d-%s', $index++, basename($entryKey)),
+                )];
+            }
+        };
+    }
+}
--- a/src/Builder/Pdf/MergePdfBuilder.php
+++ b/src/Builder/Pdf/MergePdfBuilder.php
@@ -9,7 +9,7 @@ use Sensiolabs\GotenbergBundle\Builder\Behaviors\Dependencies\AssetBaseDirFormat
 use Sensiolabs\GotenbergBundle\Builder\Behaviors\DownloadFromTrait;
 use Sensiolabs\GotenbergBundle\Builder\Behaviors\EmbedTrait;
 use Sensiolabs\GotenbergBundle\Builder\Behaviors\EncryptTrait;
-use Sensiolabs\GotenbergBundle\Builder\Behaviors\FilesTrait;
+use Sensiolabs\GotenbergBundle\Builder\Behaviors\FilesOrderTrait;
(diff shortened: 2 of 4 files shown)


src/Builder/Behaviors/FilesOrderTrait.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+namespace Sensiolabs\GotenbergBundle\Builder\Behaviors;
+
+use Psr\Log\LoggerInterface;
+use Sensiolabs\GotenbergBundle\Builder\Attributes\NormalizeGotenbergPayload;
+use Sensiolabs\GotenbergBundle\Builder\Util\NormalizerFactory;
+use Sensiolabs\GotenbergBundle\Version\Version;
+
+trait FilesOrderTrait
```
Jean-Beru, line 12: Why not add these methods to the existing trait?

src/Builder/Behaviors/FilesOrderTrait.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        $this->filesSortMode = 'call';
+
+        return $this;
+    }
+
+    #[NormalizeGotenbergPayload]
+    private function normalizeFiles(): \Generator
+    {
+        $mode = &$this->filesSortMode;
```
Jean-Beru: Using a reference is not needed here.

src/Builder/Pdf/MergePdfBuilder.php
```diff
@@ -54,6 +54,9 @@ final class MergePdfBuilder extends AbstractBuilder
         'pdf',
     ];
 
+    /**
+     * @return array<string>
```
Jean-Beru:
```suggestion
     * @return list<string>
```

src/Builder/Util/NormalizerFactory.php
```diff
@@ -219,8 +219,8 @@ public static function watermark(): \Closure
     private static function files(string $type): \Closure
     {
         return static function (string $key, array $assets) use ($type) {
-            foreach ($assets as $asset) {
-                yield [$type => new DataPart(new File($asset))];
+            foreach ($assets as $entryKey => $asset) {
+                yield [$type => new DataPart(new File($asset), basename($entryKey))];
```
Jean-Beru: Since `$asset` is an instance of `\SplFileInfo`, we can use its `getBasename` method directly.

src/Builder/Behaviors/FilesOrderTrait.php
Nitram1123: Good question, for a few reasons:
`FilesTrait` is shared across every builder that accepts files. Exposing these methods would be useless for builders other than merge.
`FilesTrait` owns the "register files" concern. Adding sort state would mix two responsibilities.
The other builders keep their current behavior with zero changes, so there's no risk of an accidental BC break.

src/Builder/Util/NormalizerFactory.php
Nitram1123: Good point in general, but here the array key carries information that $asset doesn't have: FilesOrderTrait rewrites the keys to inject the 000001- ...

Jean-Beru: Good catch.

src/Builder/Behaviors/FilesOrderTrait.php
Jean-Beru: Indeed. It's only needed by MergeBuilder.

Jean-Beru (approved):

Jean-Beru (approved):

src/Builder/Behaviors/FilesOrderTrait.php
```diff
@@ -0,0 +1,69 @@
...
+trait FilesOrderTrait
+{
+    use FilesTrait {
+        normalizeFiles as private normalizeFilesOriginal;
+    }
+
+    /**
+     * @var 'name'|'call'
+     */
+    private string $filesSortMode = 'name';
```
Jean-Beru, line 21: IMO, a string is OK since this property is purely internal. 

See [original discussion](https://github.com/sensiolabs/GotenbergBundle/pull/278#discussion_r3361531328).

Neirda24: This is a trait. So private is not private because this trait can be used by any developper. So we have to mark this property being internal but that would feel weird on a trait. WDYT ?

Neirda24: what if instead of a string key we set the callback itself. a Closure object ?

Nitram1123: A Closure would work, but we'd lose the ability to compare/assert the value cleanly (in tests, logs, etc.), and it can't be constrained to just two options — any callable becomes valid. I initially thought an enum might be overkill here, but it's probably the better option: it just stores the strategy, while keeping things type-safe and testable. What do you think?

Jean-Beru: @Nitram1123 wants an Enum
@Neirda24 wants a Callable
What about this ? :grinning: 

```php
// FilesSortMode.php
enum FilesSortMode: string
{
    case Name = 'name';
    case Call = 'call';

    public function filename(int $index, \SplFileInfo $file): string
    {
        return match ($this) {
            self::Name => $file->getFilename(),
            self::Call => \sprintf('%06d-%s', $index, $file->getFilename()),
        };
    }
}

// FilesOrderTrait.php
#[NormalizeGotenbergPayload]
private function normalizeFiles(): \Generator
{
    $mode = $this->filesSortMode;

    yield 'files' => static function (string $key, array $assets, Version $version, LoggerInterface|null $logger) use ($mode): \Generator {
        $index = 1;
        foreach ($assets as $info) {
            yield ['files' => new DataPart(new File($info), $mode->filename($index++, $info))];
        }
    };
}
```

Neirda24: as discussed with @StevenRenaux  & @Jean-Beru : we should use a Closure as property. With phpstan annotations as guards.

Neirda24 (changes requested):

PR #279, opened 2026-06-05
feat: add lazy template rendering to stream HTML at send time

Jean-Beru:

## Description

When calling `content($template, $context)` (or `header()` / `footer()`), the Twig template is rendered immediately and the resulting HTML is kept in memory until the request is sent to Gotenberg.
  
This PR adds a `$lazy` argument to `content()`, `header()` and `footer()`. When set to `true`, the template is only rendered while the HTTP request body is sent. Peak memory usage is then a single Twig output chunk instead of the full document.

Lazy rendering is disabled by default to keep BC promise.

## Example

```php
public function __invoke(GotenbergPdfInterface $gotenberg): Response
{
    return $gotenberg->html()
        ->content('invoice.html.twig', ['invoice' => $invoice], lazy: true)
        ->generate()
        ->stream();
}
```

## Screenshots

Example: 5000 lines containing an inlined svg + a paragraph.

### Before

image

### After

image

### DIFF
--- a/src/Builder/Behaviors/Chromium/ContentTrait.php
+++ b/src/Builder/Behaviors/Chromium/ContentTrait.php
@@ -13,6 +13,7 @@ use Sensiolabs\GotenbergBundle\Builder\ValueObject\StreamedPart;
 use Sensiolabs\GotenbergBundle\Enumeration\Part;
 use Sensiolabs\GotenbergBundle\Exception\PartRenderingException;
 use Sensiolabs\GotenbergBundle\NodeBuilder\ArrayNodeBuilder;
+use Sensiolabs\GotenbergBundle\NodeBuilder\BooleanNodeBuilder;
 use Sensiolabs\GotenbergBundle\NodeBuilder\ScalarNodeBuilder;
 use Sensiolabs\GotenbergBundle\Twig\GotenbergRuntime;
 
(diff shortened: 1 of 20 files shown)


src/Builder/Behaviors/Chromium/ContentTrait.php
```diff
@@ -28,14 +29,17 @@ abstract protected function getBodyBag(): BodyBag;
     /**
      * @param string               $template #Template
      * @param array<string, mixed> $context
+     * @param bool                 $lazy     When true, the template is only rendered when the HTTP request is sent,
```
Neirda24, line 32: no options to set this globally then ?

src/Builder/Behaviors/Chromium/ContentTrait.php
```diff
@@ -186,18 +196,46 @@ public function footerFile(string $path): static
...
+            }
+
+            $this->getBodyBag()->set($part->value, $renderedPart);
+
+            return $this;
+        }
+
+        // Rendering is deferred until the request body is sent so the HTML is streamed chunk
+        // by chunk to Gotenberg instead of being buffered in memory.
+        $renderer = function () use ($twig, $loadedTemplate, $template, $part, $context): \Generator {
```
Neirda24, line 212: I think we could set this variable early and then if not lazy just invoke it instead of duplicating the logic ?

src/Builder/Util/StreamedDataPart.php
```diff
@@ -0,0 +1,58 @@
...
+        private readonly \Closure $renderer,
+        string $filename,
+        string $contentType,
+    ) {
+        parent::__construct('', $filename, $contentType);
+    }
+
+    public function getBody(): string
+    {
+        return implode('', iterator_to_array(($this->renderer)(), false));
```
Neirda24, line 32: this seems to break the streamable part. Shouldn't we just throw an error instead ?

tests/Builder/Pdf/HtmlPdfBuilderTest.php
```diff
@@ -125,6 +125,35 @@ public function load(string $class): object|null
         $this->assertContentFile('index.html', 'text/html', $expected);
     }
 
+    public function testWithLazyTwigContentFile(): void
```
Neirda24, line 128: I am not sure to understand where we validate the fact that it should be deferred ?

src/Builder/Util/StreamedDataPart.php
Jean-Beru: Only `bodyToIterable` is used by `GotenbergClient`.

tests/Builder/Pdf/HtmlPdfBuilderTest.php
Jean-Beru: Indeed, it confirms that the model is rendered correctly when "lazy" mode is enabled. Only `StreamedDataPartTest` tests check chunks.

src/Builder/Behaviors/Chromium/ContentTrait.php
Jean-Beru: Are you thinking of a `->lazy()` method with the relating semantic configuration property ?

src/Builder/Behaviors/Chromium/ContentTrait.php
Jean-Beru: Good catch!

src/Builder/Behaviors/Chromium/ContentTrait.php
Neirda24: yes please :)

src/Builder/Util/StreamedDataPart.php
Neirda24: As discussed: LogicException

PR #286, opened 2026-07-13
Fix content encoding issue that arises when using a reverse proxy

yard-mschwartz:
| Issues                  | N/A |

## Description

This is a fix for a problem I ran into earlier today, namely Firefox showing me:

```
Content Encoding Error

The page you are trying to view cannot be shown because it uses an invalid or unsupported form of compression.
```

The cause:

We configured an external Gotenberg server behind a reverse proxy (long story as to why). Symfony tells this server to gzip the response and decompresses the PDF when it receives it back from the Gotenberg server. So far so good.

Then, when I stream the response back to the client, the response header still contains the metadata of the gzipped response, which is wrong. The `content-encoding` and `content-length` of the gzipped response is not the same as those of the streamed response.

```
#[Route('/{myEntity}/pdf', methods: ['GET'])]
public function pdf(Entity $myEntity): StreamedResponse
{
    // more logic here

    return $this->gotenberg
        ->html()
        ->content(...)
        ->generate()
        ->stream();
}
```

The workaround to add `Accept-Encoding` also worked for me:

```
framework:
    http_client:
        scoped_clients:
            gotenberg.client:
                base_uri: '%env(GOTENBERG_DSN)%'
                headers:
                    Accept-Encoding: 'identity'
```

But it relies on the proxy server honoring this setting.

The fix removes these two `content-*` headers, because `Content-Encoding: gzip` is wrong for the streamed response and `Content-Length` is irrelevant for a streamed response.

### DIFF
--- a/src/Builder/Result/GotenbergFileResult.php
+++ b/src/Builder/Result/GotenbergFileResult.php
@@ -114,6 +114,7 @@ class GotenbergFileResult extends AbstractGotenbergResult
         $filename = $this->getFileName();
 
         $headers = $this->getHeaders();
+        unset($headers['content-encoding'], $headers['content-length'], $headers['transfer-encoding'], $headers['connection']);
         $headers['X-Accel-Buffering'] = ['no']; // See https://symfony.com/doc/current/components/http_foundation.html#streaming-a-json-response
         if ($filename) {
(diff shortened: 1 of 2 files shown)


Neirda24:
thank you for this. @yard-mschwartz

Neirda24:
failing tests are unrelated

PR #280, opened 2026-06-08
chore(ci): fix PHPStan tests

Jean-Beru:
| Q                       | A        |
|-------------------------|----------|
| Gotenberg API version ? | 8.x      |
| Bug fix ?               | no   |
| New feature ?           | no   |
| BC break ?              | no   |
| Issues                  |  |

## Description

Fix the latest PHPStan errors which are no longer reported in each class but in the `WebhookTrait`.

PR #288, opened 2026-07-23
fix: repair PHPStan static analysis (stale ignores + WebhookTrait narrowing)

marilenaRM:
## What

The **PHPStan** static-analysis job on `1.x` is currently red. It is not a code regression — it's dependency drift: the job runs on PHP 8.4 / Symfony 7.3 with a recent PHPStan whose stricter (and, across patch releases, shifting) type narrowing changed which errors are reported.

Two categories of failure:

- **Stale builder ignores.** The `is_array()` / `is_string()` "will always evaluate to true" `ignoreErrors` entries for every PDF and Screenshot builder no longer match anything, so PHPStan reports them as non-ignorable `ignore.unmatched`.
- **`WebhookTrait::webhookConfigurationValidator()`.** The defensive checks on `method` / `route` are proven redundant because the `WebhookConfiguration` PHPDoc already narrows those values. The exact error even *moves between runs* as dependencies get patched (`function.alreadyNarrowedType` on `in_array`/`is_string`/`is_array`, or `booleanAnd.alwaysFalse` on the `&&`), so a message- or line-specific ignore can't track it.

## Fix

Config-only, no application-code change:

- Remove the obsolete `is_array()` / `is_string()` builder ignore entries.
- Set `reportUnmatchedIgnoredErrors: false`, since these PHPDoc-narrowing diagnostics vary across dependency patch releases.
- Add two identifier-based ignores for `WebhookTrait` (`function.alreadyNarrowedType` + `booleanAnd.alwaysFalse`) so the validator's defensive checks stay covered whichever variant a given run produces.

The defensive validation in `WebhookTrait` is intentionally kept as-is; happy to instead drop the now-dead branches if you'd prefer removing them over ignoring.

### DIFF
--- a/phpstan.dist.neon
+++ b/phpstan.dist.neon
@@ -9,6 +9,7 @@ parameters:
 	bootstrapFiles:
 		- phpstan-bootstrap.php
 	level: 8
+	reportUnmatchedIgnoredErrors: false
 	paths:
 		- 'config'
 		- 'src'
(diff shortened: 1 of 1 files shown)


Jean-Beru:
Thanks for the fix but I think this duplicates https://github.com/sensiolabs/GotenbergBundle/pull/280.

marilenaRM:
> Thanks for the fix but I think this duplicates #280.

ok then I close this one, sorry!

PR #289, opened 2026-08-21
fix(docs): keep @example tags when a method is seen through several traits

marilenaRM:
| Issues                  | -        |

## Description

`BuilderParser::prepareBuilderFromClass()` walks parents, interfaces and traits, then the class
itself, merging what it finds along the way. The merge branch propagated `package`, `description`,
`param` and `see` — but not `example`.

So as soon as a method was reached a second time (which happens for every method a builder gets
through an aggregate trait such as `ChromiumPdfTrait`), its `@example` tag was replaced by the empty
one coming from the later pass, and the code sample silently disappeared from the generated
documentation.

The six Chromium builders were affected: `addAsset()` lost its sample in all of them. Regenerating
brings it back.

The two `AssetTrait` samples that come back were not valid PHP — `__DIR__'/../../public/admin.jpeg'`
is a syntax error — and the `addAsset()` one passed two paths where the method takes a single one
(`addAsset(string|\Stringable $path)`). Both are fixed here, otherwise this PR would put broken code
back into the docs.

## Test plan

- `php ./docs/generate.php` — the only diff is the six restored samples, nothing else moves
- `./vendor/bin/phpunit` — 1095 tests, 2217 assertions, green
- `./vendor/bin/php-cs-fixer check` — clean

### DIFF
--- a/docs/generate.php
+++ b/docs/generate.php
@@ -361,6 +361,10 @@ class BuilderParser
                     $parsedDocBlock['tags']['see'],
                 )));
             }
+
+            if (isset($parsedDocBlock['tags']['example']) && [] !== $parsedDocBlock['tags']['example']) {
+                $this->parts['methods']['@'][$method->getShortName()]['tags']['example'] = $parsedDocBlock['tags']['example'];
+            }
         }
     }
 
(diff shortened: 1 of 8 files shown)


Neirda24 (approved):

Jean-Beru (approved):

Jean-Beru:
Thanks @marilenaRM !

PR #290, opened 2026-08-21
Split ContentTrait into PDF and screenshot flavours, deprecate what the routes ignore

marilenaRM:
| Bug fix ?               | yes + refactor |
| Issues                  | Refs #257 |

## Description

Supersedes #274, which took a different route. Opening a fresh PR because the approach changed
entirely after the review — keeping the old discussion attached to code that no longer exists
would have been confusing.

### The problem

Chromium PDF routes and Chromium screenshot routes do not read the same body parts, but both
families shared a single `ContentTrait`:

| Route | `index.html` | `header.html` / `footer.html` |
|---|---|---|
| `convert/url` | ignored (body comes from the URL) | read |
| `convert/html`, `convert/markdown` | read | read |
| `screenshot/url` | ignored | ignored |
| `screenshot/html`, `screenshot/markdown` | read | ignored |

So `UrlPdfBuilder::content*()` and every `header*()` / `footer*()` on the screenshot builders send
parts that Gotenberg silently drops.

### The approach

We already split `PageProperties` into `PdfPagePropertiesTrait` and
`ScreenshotPagePropertiesTrait`. Same situation here, so the same treatment:

* `Chromium\PdfContentTrait` — `content*()` + `header*()` / `footer*()`, a drop-in replacement for
  the old trait;
* `Chromium\ScreenshotContentTrait` — same surface, but `header*()` / `footer*()` are deprecated;
* `Chromium\PartTrait` (`@internal`) — the `withRenderedPart()` / `withRawPart()` / `withFilePart()`
  plumbing both need. Extracted rather than duplicated: two copies of the Twig rendering and path
  resolution would drift.
* `Chromium\ContentTrait` — deprecated, reduced to `use PdfContentTrait;` with a file-level
  `trigger_deprecation()` so it only fires for third parties still using it.

No `insteadof` anywhere, and the aggregates just swap one `use` line each.

### Nothing changes at runtime

This is the part worth reviewing closely. Every deprecated method keeps its previous behaviour and
the multipart payload is byte-identical:

* `ScreenshotContentTrait::normalizeContent()` still yields `header.html` and `footer.html`, in the
  same order `ContentTrait` did, so the multipart parts come out identical. Dropping them would have
  turned today's silently-ignored part into a `FormDataPart` exception, because the raw
  `RenderedPart` would fall back to `NormalizerFactory::noop()`. A deprecation must not change
  behaviour.
* The `#[WithConfigurationNode]` attributes stay on the deprecated `header()` / `footer()`, so the
  `default_options.screenshot.*` config keys keep working. They go through `BuilderConfigurator`,
  which calls the method, so a YAML user gets the deprecation too — covered by
  `BuilderConfiguratorTest`.

  Note the config node itself is *not* flagged deprecated, so `config:dump-reference` says nothing:
  `src/NodeBuilder/` has no equivalent of Symfony's `->setDeprecated()`. Adding one felt like a
  separate PR — happy to do it here if you'd rather.
* `content*()` on the URL builders still renders and sends `index.html`, exactly as before.

### 2.0 is a pure deletion

* delete `ContentTrait.php`;
* delete the six `header*()` / `footer*()` from `ScreenshotContentTrait` and its two extra yields;
* delete the three `content*()` from each URL builder — the aggregates already alias them
  `as private`, so removing the class-level overrides hides them instead of re-exposing them.

That last alias is the reason the URL builders declare the deprecated methods in the class rather
than in a dedicated trait: a class method wins over a trait method, so no `insteadof` is needed, and
the `as private` block guarantees the 2.0 pass cannot accidentally bring them back.

### Also in here

* `symfony/deprecation-contracts` added to `require` — `#[\Deprecated]` only fires at runtime on
  PHP 8.4+, and we still support 8.1;
* `UPGRADE-1.5.md`, 1.5 being the next minor after v1.4.0;
* `docs/generate.php` now renders a `> [!WARNING]` block for `@deprecated` methods (and drops their
  `@example`) instead of leaving them undocumented.

## Test plan

- `./vendor/bin/phpunit` — 1104 tests, 2250 assertions, green. Tests added: the trait-level
  deprecation; `header*()` / `footer*()` deprecated but still sent, as a shared test-case trait so
  the three screenshot builders are all covered; `content*()` still sent on both URL builders; the
  `default_options.screenshot.*` config path; and a negative check that PDF builders trigger
  nothing.
- `./vendor/bin/php-cs-fixer check` — clean
- `./vendor/bin/composer-dependency-analyser` — no issues
- `php ./docs/generate.php` — leaves the tree clean, the committed docs are what the generator produces


src/Builder/Behaviors/Chromium/PdfContentTrait.php
```diff
@@ -0,0 +1,182 @@
...
+
+    /**
+     * @param string               $template #Template
+     * @param array<string, mixed> $context
+     *
+     * @throws PartRenderingException if the template could not be rendered
+     *
+     * @example content('content.html.twig', ['my_var' => 'value'])
+     */
+    public function content(string $template, array $context = []): static
```
Jean-Beru: It is a BC break since previous declaration in `ContentTrait` is `public function content(string $template, array $context = []): self`.

We could fix all return type hints in a dedicated PR in 2.0.

marilenaRM: Confirmed, and it is worse than a type-hint nitpick: a third-party **non-final** builder using `ContentTrait` whose subclass overrides `content(): self` compiles on 1.4 but fatals on this branch — PHP only accepts `static` as an override of `static`, and since the deprecated `ContentTrait` now delegates to `PdfContentTrait`, even untouched legacy code picks up the new signature at class-link time. Our own builders are all `final`, which is why neither PHPUnit nor PHPStan caught it.

Fixed in b697d2a: `content()`, `contentRaw()` and `contentFile()` are back to `: self` in both traits and in the URL builders' deprecated overrides. `header*()`/`footer*()` already returned `static` in 1.4, so they are left as is. Agreed on doing the wholesale `self` → `static` pass in a dedicated PR for 2.0.

tests/Builder/Behaviors/Chromium/ContentTraitTest.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace Sensiolabs\GotenbergBundle\Tests\Builder\Behaviors\Chromium;
+
+use PHPUnit\Framework\Attributes\PreserveGlobalState;
+use PHPUnit\Framework\Attributes\RunInSeparateProcess;
+use PHPUnit\Framework\TestCase;
+use Sensiolabs\GotenbergBundle\Tests\CollectDeprecationsTrait;
+
+final class ContentTraitTest extends TestCase
```
Jean-Beru, line 10: IMO, there is no need to test the deprecation trigger here. It adds too much complexity for something that will be removed in the next major version.

tests/CollectDeprecationsTrait.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace Sensiolabs\GotenbergBundle\Tests;
+
+trait CollectDeprecationsTrait
```
Jean-Beru, line 5: If merged as it, we must replace it by `expectUserDeprecationMessage` in 2.0.

(not possible yet because PHPUnit 11 needs PHP 8.2 needs SF7.0)

PR #292, opened 2026-08-28
fix: reconcile the config mapping callback type between BuilderStack and BuilderConfigurator

### LINKED ISSUE #291: [Bug] BuilderStack and BuilderConfigurator declare incompatible types for the same config mapping
## Description

`BuilderStack` and `BuilderConfigurator` describe the very same array — the config mapping the
extension dumps into the container — with two incompatible PHPStan types for its `callback` entry.

`src/DependencyInjection/BuilderStack.php` — `$configMapping` and `getConfigMapping()`:

```
'callback': array<array-key, string>|null
```

`src/Configurator/BuilderConfigurator.php` — `__construct()` `$configurations`:

```
'callback': (\Closure(mixed): mixed)|null
```

The `\Closure` one is the wrong one. `BuilderStack` stores array callables:

```php
$callback = [$attribute->node->enumClass, 'from'];   // NativeEnumNodeBuilder
$callback = [Unit::class, 'parse'];                  // UnitNodeBuilder
```

and a closure could never reach `BuilderConfigurator` in the first place, because the mapping goes
through `Definition::replaceArgument()` into the compiled container:

```
Symfony\Component\DependencyInjection\Exception\RuntimeException:
Unable to dump a service container if a parameter is an object or a resource, got "Closure".
```

## Why PHPStan does not catch it

The two types only meet at compile time, in `SensiolabsGotenbergExtension`:

```php
$container->getDefinition('sensiolabs_gotenberg.builder_configurator')
    ->replaceArgument(0, $this->builderStack->getConfigMapping())
```

`replaceArgument()` takes `mixed`, so the two shapes never meet in analysed code. Connect them
directly and the error shows up straight away:

```php
$builderStack = new BuilderStack();
$builderStack->push(UrlPdfBuilder::class);

new BuilderConfigurator($builderStack->getConfigMapping(), []);
```

```
Parameter #1 $configurations of class BuilderConfigurator constructor expects
  array<..., array{method: string, mustUseVariadic: bool, callback: (Closure(mixed): mixed)|null}>,
  array<..., array{method: string, mustUseVariadic: bool, callback: array<string>|null}> given.
  argument.type
  Offset 'callback' ((Closure(mixed): mixed)|null) does not accept type array<string>|null.
```

## Impact

Nothing is broken at runtime: `$configurationMap['callback']($value)` works fine for array
callables, and enum / `Unit` config values are converted correctly. This is a type-safety defect —
anyone trusting the `\Closure` docblock (a test, a decorator, a future refactor) is writing against
a contract the container is physically unable to honour.

## Note for whoever picks this up

Rewriting both sides to `(callable(mixed): mixed)|null` does **not** work. PHPStan collapses the two
assignments into `array{class-string<BackedEnum>, 'from'|'parse'}` and refuses it as a callable,
since `parse` is not declared on `\BackedEnum`. Same outcome for `callable-array`, for an
intersection type, and for a union of two precise tuples — PHPStan re-collapses that one at the call
site. The fix has to make the declared type honest about the array callable rather than dress it up
as a closure.

Spotted while writing `BuilderConfiguratorTest` in #290, which currently sidesteps the problem by
hand-writing the mapping literal instead of building it with `BuilderStack`.

marilenaRM:
| Issues                  | Fix #291 |

## Description

`BuilderStack` and `BuilderConfigurator` described the same array — the config mapping the extension
dumps into the container — with two incompatible types for its `callback` entry: `array<array-key,
string>|null` on one side, `(\Closure(mixed): mixed)|null` on the other. See #291 for the full
write-up.

### The `\Closure` side was the lie

`BuilderStack` stores array callables (`[$enumClass, 'from']`, `[Unit::class, 'parse']`), and a
closure could not reach `BuilderConfigurator` even if we wanted it to — the mapping goes through
`Definition::replaceArgument()` into the compiled container, and `PhpDumper` refuses:

```
RuntimeException: Unable to dump a service container if a parameter is an object or a resource, got "Closure".
```

I checked this before picking a direction, since "build real closures in `BuilderStack`" is the
obvious first idea and it is a dead end.

### Why not just write `callable(mixed): mixed` on both sides

PHPStan collapses the two assignments into `array{class-string<BackedEnum>, 'from'|'parse'}` and
rejects it as a callable, because `parse` is not declared on `\BackedEnum`. I also tried
`callable-array`, an intersection `array{class-string<\BackedEnum>, string}&callable(mixed): mixed`,
per-branch assignment to dodge the variable merge, and a union of two precise tuples. The union one
is the instructive failure: the assignments pass, then PHPStan re-collapses the type at the call site
and reports `Trying to invoke array{class-string<BackedEnum>, 'from'|'parse'} but it might not be a
callable`. PHPStan will not prove these array literals callable, whatever you declare.

### What this PR does

Both sides now share a single `@phpstan-type BuilderConfigurationMapping` declared on
`BuilderConfigurator` and imported by `BuilderStack` — same producer/consumer idiom already used for
`WebhookDefinition` in `WebhookConfigurationRegistryInterface`. One declaration, so they cannot drift
apart again.

The callback is typed `array{class-string<\BackedEnum>, string}|null`, which is exactly what ends up
in the compiled container:

```php
'pdf_format' => ['method' => 'pdfFormat', 'mustUseVariadic' => false,
                 'callback' => ['Sensiolabs\\GotenbergBundle\\Enumeration\\PdfFormat', 'from']]
```

(read out of the generated `getSensiolabsGotenberg_BuilderConfiguratorService.php` after booting the
test kernel, to confirm the shape survives dumping.)

The call site destructures instead of invoking, so PHPStan can follow the static call without an
`assert()`, an inline `@var` or an ignore:

```php
[$callbackClass, $callbackMethod] = $configurationMap['callback'];
$value = $callbackClass::$callbackMethod($value);
```

Behaviour is identical — PHP resolved `[$class, $method]($value)` to the same static call.

I considered replacing the callback with a discriminator enum plus a `match` dispatch. Rejected: it
forces `$value` (genuinely `mixed`) to be narrowed at the call site, which means adding validation
that changes behaviour. Not what a type-hint fix should do.

### Test

`tests/Configurator/BuilderConfiguratorTest.php` builds the mapping with `BuilderStack` and feeds it
straight into `BuilderConfigurator`, so the two shapes now meet in analysed code and a regression
becomes a PHPStan error rather than a silent divergence. It covers both conversions — enum
(`pdf_format`, `emulated_media_type`) and `Unit` (`paper_width`, `margin_top`, including the variadic
spread). I mutation-checked it by neutering the callback line: both tests fail with a `TypeError`.

## Checks

Run on `1.x` (8ceaa92) with PHP 8.3 / Symfony 7.4:

- `./vendor/bin/phpunit` — 1097 tests, 2221 assertions, OK
- `./vendor/bin/composer-dependency-analyser` — no issues
- `php-cs-fixer` — clean

### DIFF
--- a/src/Configurator/BuilderConfigurator.php
+++ b/src/Configurator/BuilderConfigurator.php
@@ -4,11 +4,21 @@ namespace Sensiolabs\GotenbergBundle\Configurator;
 
 use Sensiolabs\GotenbergBundle\Builder\BuilderInterface;
 
+/**
+ * The mapping is built at compile time by {@see \Sensiolabs\GotenbergBundle\DependencyInjection\BuilderStack} and
+ * dumped into the container, so a callback can only be a static method target — the container cannot dump a closure.
+ *
+ * @phpstan-type BuilderConfigurationMapping array<class-string<BuilderInterface>, array<string, array{
+ *     method: string,
+ *     mustUseVariadic: bool,
+ *     callback: array{class-string<\BackedEnum>, string}|null,
+ * }>>
+ */
 final class BuilderConfigurator
 {
     /**
-     * @param array<class-string<BuilderInterface>, array<string, array{'method': string, 'mustUseVariadic': bool, 'callback': (\Closure(mixed): mixed)|null}>> $configurations
-     * @param array<class-string<BuilderInterface>, array<string, mixed>>                                                                                       $values
+     * @param BuilderConfigurationMapping                                 $configurations
+     * @param array<class-string<BuilderInterface>, array<string, mixed>> $values
      */
     public function __construct(
         private readonly array $configurations,
@@ -28,7 +38,8 @@ final class BuilderConfigurator
             }
 
             if (null !== $configurationMap['callback']) {
-                $value = $configurationMap['callback']($value);
+                [$callbackClass, $callbackMethod] = $configurationMap['callback'];
+                $value = $callbackClass::$callbackMethod($value);
             }
 
             if (\is_array($value) && true === $configurationMap['mustUseVariadic']) {
--- a/src/DependencyInjection/BuilderStack.php
+++ b/src/DependencyInjection/BuilderStack.php
@@ -5,6 +5,7 @@ namespace Sensiolabs\GotenbergBundle\DependencyInjection;
 use Sensiolabs\GotenbergBundle\Builder\Attributes\WithBuilderConfiguration;
 use Sensiolabs\GotenbergBundle\Builder\Attributes\WithConfigurationNode;
 use Sensiolabs\GotenbergBundle\Builder\BuilderInterface;
+use Sensiolabs\GotenbergBundle\Configurator\BuilderConfigurator;
 use Sensiolabs\GotenbergBundle\Enumeration\Unit;
 use Sensiolabs\GotenbergBundle\NodeBuilder\ArrayNodeBuilder;
 use Sensiolabs\GotenbergBundle\NodeBuilder\NativeEnumNodeBuilder;
@@ -12,6 +13,8 @@ use Sensiolabs\GotenbergBundle\NodeBuilder\NodeBuilderInterface;
 use Sensiolabs\GotenbergBundle\NodeBuilder\UnitNodeBuilder;
 
 /**
+ * @phpstan-import-type BuilderConfigurationMapping from BuilderConfigurator
+ *
  * @internal
  */
 final class BuilderStack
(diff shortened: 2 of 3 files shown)


Jean-Beru (approved):
Good catch! Thank you @marilenaRM

StevenRenaux (approved):

Jean-Beru:
Thanks @marilenaRM

PR #297, opened 2026-09-16
feat: Add facturX builder + fields (release 8.34)

StevenRenaux:
| Gotenberg API version ? | 8.34      |
| BC break ?              | yes|

## Description

Related to https://github.com/gotenberg/gotenberg/releases/tag/v8.34.0
Add Factur-X/ZUGFeRD support on Chromium, LibreOffice, Merge and Split builders and add a new standalone FacturXPdfBuilder for the dedicated route


src/Builder/Behaviors/FacturXTrait.php
```diff
@@ -0,0 +1,129 @@
...
+trait FacturXTrait
+{
+    use AssetBaseDirFormatterAwareTrait;
+    use LoggerAwareTrait;
+
+    abstract protected function getBodyBag(): BodyBag;
+
+    /**
+     * The Factur-X/ZUGFeRD CII invoice XML, embedded as `factur-x.xml` regardless of the
+     * uploaded filename. Must be provided together with `facturxConformanceLevel()`.
```
Neirda24: It seems to require the call to that method either way. Maybe as an argument to this method instead?

src/Enumeration/FacturXConformanceLevel.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Sensiolabs\GotenbergBundle\Enumeration;
+
+enum FacturXConformanceLevel: string
```
Neirda24: Do we have a link to provide to understand those values ?

src/Builder/Behaviors/FacturXTrait.php
```diff
@@ -0,0 +1,129 @@
...
+     * As an asset file, by default the file is fetched in the assets folder
+     * of your application. For more information about path resolution go to
+     * assets documentation.
+     *
+     * @see https://gotenberg.dev/docs/manipulate-pdfs/factur-x
+     *
+     * @example facturxXml('invoice.xml')
+     */
+    #[WithConfigurationNode(new ScalarNodeBuilder('facturx_xml'))]
+    public function facturxXml(string|\Stringable $path): self
```
Jean-Beru: We could also accept `\SplFileInfo` if people want to bypass the path resolution.

src/Builder/Behaviors/FacturXTrait.php
```diff
@@ -0,0 +1,129 @@
...
+     *
+     * As an asset file, by default the file is fetched in the assets folder
+     * of your application. For more information about path resolution go to
+     * assets documentation.
+     *
+     * @see https://gotenberg.dev/docs/manipulate-pdfs/factur-x
+     *
+     * @example facturxXml('invoice.xml')
+     */
+    #[WithConfigurationNode(new ScalarNodeBuilder('facturx_xml'))]
```
Jean-Beru: IMO, this configuration node is not needed since we usually don't have a single `invoice.xml` to embed.

src/DependencyInjection/SensiolabsGotenbergExtension.php
```diff
@@ -38,7 +38,8 @@
  *              encrypt: array<string, mixed>,
  *              embed: array<string, mixed>,
  *              stamp: array<string, mixed>,
- *              watermark: array<string, mixed>
+ *              watermark: array<string, mixed>,
+ *              factur_x: array<string, mixed>
```
Jean-Beru:
```suggestion
 *              factur_x: array<string, mixed>,
```

